PR #35, opened 2012-04-26
View and jump to PDF in Linux with Okular

lklein:
Thanks for the great tool!  I use it on Mac and Windows.  Now I just added some simple implementation to add these functions to I can use it on Ubuntu, too.


ig0774:
Okular support is now included in the latest release

PR #41, opened 2012-06-06
Uses Preview.app if Skim.app is not available

rsobik:
This commit adds a Preview.app fallback if Skim.app is not available to view the generated PDF.


r-stein:
I close this, because it is not compatible with the current implementation.

Issue #10, opened 2011-12-11
biblatex and biber


rakke:
Hi Marciano, 
First of all I would like to thank you for a very good package.

I was wondering if you had any plans of supporting biblatex and biber?
It would be greatly appreciated as these packages adds utf8 capabilities, and hence, in many cases are a better option than bibtex. 

cheers,
Jørgen

codethief:
For me, compiling using XeTeX+biber+biblatex and auto-complete upon \cite{]ing work out of the box. I'm running LaTeXtools v2014.10.19.05.12.13 with the default config, together with latexmk 4.39 (from a manually installed TeXlive 2013 distribution on Ubuntu 12.04 and without any ~/.latexmkrc file).

The reason this works is that, [as of version 4.22](http://users.phys.psu.edu/~collins/latexmk/versions.html), latexmk automatically recognizes whether to use biber (as @fnurl already pointed out). Hence it doesn't make any difference for LaTeXtools whether you use bibtex or biber because it only cares about your .bib files (whose format does not depend on your choice of bibtex/biber).

So, from my point of view, this issue can be closed, except for custom builders not using latexmk. But it might be better then to either rename this issue or create a new one.

@harishkumarholla: I've added an [answer](http://tex.stackexchange.com/a/220501/26566) to the question on stackexchange. Feel free to add your instructions regarding custom builders.

Nhapsie:
I am a newbie, but use biblatex and biber without a problem with this on top of the preamble:
%-*- program: xelatex -*-
%-*- program: biber -*-
%-*- program: xelatex -*-

msiniscalchi:
When #617 is merged, this will provide another way to achieve the same objective. Closing.

PR #42, opened 2012-06-07
Issue #10 - biblatex support

ssalonen:
Minimal support for \addbibresource{} and for biblatex


ig0774:
Closing this as we already have `\addbibresource`. Support for more biblatex commands is on the way.

PR #52, opened 2012-07-15
Added a trigger to create german question marks

jbumueller:
This is great sublime bundle thanks a lot for that. 

I really like to use the triggers, here is a new one: gl [tab] creates `\glqq \grqq` with the first tab it jums between them and with the second it jumps after them.


r-stein:
I close this, because you usually use babel with ``"` `` and `"'` for german quotes.

PR #88, opened 2012-10-11
Portable setup of Sumatra and additional env variables

williamledoux:
I needed to have a portable folder that include ST2, MikTex and Sumatra, compiles with xelatex, and that will work anywhere without having anything on the path, any specific environment variable, or any default PDF viewer.

* I added the windows specific 'sumatra' parameter in the preference file, so that you can setup globally the location of the viewer

* I added the 'env' parameter to the build system so that you can customize compilation, for instance, changing the pdflatex engine or set up the latex input paths. I did not put this in the preference file because one may want to have multiple build systems with different latex engine or latex inputs.

Here is an example of how to use this to customize a miktex build system on windows:

    "env": {
        "PDFLATEX":"xelatex",
        "TEXINPUTS":"Q:/@Administration/@Include;Q:/@Documents/@Include;./@Include;.",
        "BIBINPUTS":"Q:/@Documents/@Bibs"
    },


williamledoux:
I now have my builder (forked from traditonnal builder) that does this, so I close the pull request.
Most of my changes would fit in the traditional builder, do you want a pull request for this ?

msiniscalchi:
Yes please.

williamledoux:
Cool :)
There is just one thing haunting me, that I'd like to have your point of view about.
I am modifing `os.environ` inside the builder with some dictionary that is set in the `builder_settings`.
The thing is, if one of the builder's command fails, I won't be able to restore the previous environment. So what do you think:

1. It's not a big deal
2. Additionnal env variables should be read in platform settings and handled in makePDF.py so that it can be restored (the same way the PATH variable is).
3. Something else ?

msiniscalchi:
I vote for #2, assuming all the variables you set are build-related.
Actually, in this case, I think the proper place in the settings file is
among the *builder* settings (possibly platform-specific:i.e.,
"builder_settings", "windows".

On Wed, Mar 18, 2015 at 8:41 AM, William Ledoux <notifications@github.com>
wrote:

> Cool :)
> There is just one thing haunting me, that I'd like to have your point of
> view about.
> I am modifing os.environ inside the builder with some dictionary that is
> set in the builder_settings.
> The thing is, if one of the builder's command fails, I won't be able to
> restore the previous environment. So what do you think:
>
>    1. It's not a big deal
>    2. Additionnal env variables should be read in platform settings and
>    handled in makePDF.py so that it can be restored (the same way the PATH
>    variable is).
>    3. Something else ?
>
> —
> <https://github.com/SublimeText/LaTeXTools/pull/88#issuecomment-82976525>.
>



-- 
Marciano Siniscalchi

PR #144, opened 2013-02-06
Added the ability to easily change environment and specify a default bibtex bibliography

jhwilson:
# Ability to change environment

By pressing `C-l,u`, Sublime prompts the user for a new environment name, and it replaces the current environment name with the user-specified one (to easily change `equation` to `eqnarray`, `align` to `multline`, etc.).
# Default bibliography

Added a new variable to specify a default bibliography in

`LaTeXTools Preferences.sublime-settings`

The variable is called `default_bibliography` and defaults to `false` (i.e. no default bibliography specified). If a `default_bibliography` is specified, it takes as input the full path to a .bib file the user wants to use by default.

The default will be used by `latex_cite_completions.py` when (1) no `.bib` file is referenced by a `\bibliography` command, or (2) Any `.bib` file fails to open (e.g. if you have `\bibliography{refs}` in your `.tex` file, but `refs.bib` does not exist in the same directory as your `.tex` file). This was added to both the completions and the command.

This does seem to conflict with other pull requests. I'll be happy to make any changes needed to have this work with those.

Lastly, I apologize for what may be sloppy python -- I'm learning as I'm going. Any pointers would be appreciated.


r-stein:
I close this, because it is not compatible and we got the change env command in the meantime. Not sure whether we should support the default bib feature, but it should be in a separate PR.

PR #198, opened 2013-06-21
Errors might happen in classes and not in files

GhostGambler:
This is a fix to correctly display the source of an error found in the log file.


ig0774:
I'm closing this since there isn't enough information to determine whether or not this is still necessary (and I don't want tochange the log parsing code unless we're really sure it's necessary).

PR #260, opened 2013-08-27
Added additional backup types for {journal} in the cite_panel_format.

andrewning:
I like to use this format:

```
"cite_panel_format": ["{author_short} {year} - {title_short} ({keyword})","{journal}"]
```

rather than the default:

```
"cite_panel_format": ["{author_short} {year} - {title_short} ({keyword})","{title}"]
```

because I think it's redundant to show the title at the top and the title again just below.  Of course this capability is already provided but the "{journal}" type only includes journal articles and so doesn't show any relevant information  for other types.  I added a few more types as described below.
- Just like {author} already falls back on editor for books, {journal} now refers to the journal name for articles but falls back to: booktitle (conferences), institution (tech reports), publisher (books), and school (thesis or dissertation).
- Also simplified some of the code relating to this change so that the same code wasn't being called in multiple places and it would be easier to maintain without introducing errors.
- Other lines are highlighted because my editor auto-removed trailing whitespace

### DIFF
--- a/latex_cite_completions.py
+++ b/latex_cite_completions.py
@@ -1,5 +1,5 @@
 # ST2/ST3 compat
-from __future__ import print_function 
+from __future__ import print_function
 import sublime
 if sublime.version() < '3000':
     # we are on ST2 and Python 2.X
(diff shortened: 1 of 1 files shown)


r-stein:
I close this, because I think it has been implemented elsewhere in the meantime.

PR #268, opened 2013-09-11
Show building in status bar, show build panel follows preference

Azhag:
Show the building process in the status bar. Depends on Show_build_status parameter, set in LatexTools settings.

Can now disable automatic popup of build results, based on the parameter set in global settings or LatexTools settings. Combines well with forked BuildOnSave, allowing for hidden build panel, appearing only when errors are present during the building process.

(cleaned indentation as well)


andrejbauer:
It's ridicolous that an issue like this is open for half a year. It's even more ridicolous that LaTeXtools does ignores `show_panel_on_build`.

ig0774:
Thanks for the code. Something much like this has been added in v3.11.0. We don't actually respect the `show_panel_on_build` setting mostly because we provide a `hide_build_panel` setting that offers more fine-grained control over whether the build panel is shown or not.

PR #270, opened 2013-09-18
Update README.markdown

alexandregme:
I've got a lot of problems trying to compile the text file I think it will be ok, if you tell the user to make sure that he put the miktex/bin at the environment.


r-stein:
I close this, because it is not compatible with the current README.

PR #275, opened 2013-10-06
Support relative specification of TEXroot in project file in addition to existing behaviour

cpbotha:
This is a small change that enables users to specify TEXroot in their project files relative to the tex file that is being edited. The old behaviour remains unchanged, so existing users won't notice the difference.

I've also added a paragraph to the README documenting TEXroot specification in the project file. This was a todo.


ig0774:
I hadn't seen this before and it actually looks kinda nice. Unfortunately, I just merge #629, which kind of conflicts with the purpose of this PR. For posterities' sake, let me just say that determining the `TEXroot` relative to the project file is less ambiguous in projects that span multiple folders than those which have all the `tex` files in a single folder. For the latter case, I think #629 accomplishes the same goal (provided that the `.sublime_project` file is also in the same folder).

PR #289, opened 2013-11-18
Update delete_temp_files.py to remove .nav and .snm files

monty5811:
'.nav' and '.snm' files are created when working with beamer documents. 

Added these extensions to the temp_exts list for deletion.


ig0774:
I think this is pretty much covered by #520. Thanks though!

PR #295, opened 2013-12-06
fix jump to pdf from command palette

chid:
Currently when calling Jump to PDF from the command palette in ST3 it has this error (and nothing happens)

```
  File "C:\Program Files (x86)\Sublime Text 3\sublime_plugin.py", line 549, in run_
    return self.run(edit)
  File "AppData\Roaming\Sublime Text 3\Packages\LaTeXTools\jumpToPDF.py", line 36, in run
    from_keybinding = args["from_keybinding"]
KeyError: 'from_keybinding'
```

This commit fixes that.


ig0774:
Thanks for this. We've got a similar, albeit more verbose and hacky fix in `master` already.

PR #307, opened 2014-01-01
Fixes undefined function call to raw_input() (py3)

### LINKED ISSUE #306: the latex parser has an undefined function
where is this function 
https://github.com/SublimeText/LaTeXTools/blob/master/parseTeXlog.py#L82
defined in the source code? seems to me to be coming out of thin air

chid:
Fixes #306


ig0774:
Thanks. This fix has been integrated.

PR #315, opened 2014-01-24
Support showing the eprint bibtex entry in the cite panel

aleth:
This adds support for an "eprint" keyword for settings such as cite_panel_format, so that the bibtex eprint entry can be displayed in the cite panel etc. The bibtex entry is wrapped in angular brackets [...] if it is not empty, just as they are often displayed in publications (doing this by hand in cite_panel_format would lead to many [] for all those bibtex entries that don't have an eprint associated with them).

### DIFF
--- a/latex_cite_completions.py
+++ b/latex_cite_completions.py
@@ -244,6 +244,7 @@ def get_cite_completions(view, point, autocompleting=False):
         authors = []
         years = []
         journals = []
+        eprints = []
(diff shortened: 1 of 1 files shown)


ig0774:
We have support for "eprint", though not in quite the same way. I might look into making the formatting engine a bit more... robust to handle these sorts of things. Thanks for this though!

PR #317, opened 2014-02-01
Fixed cite autocompletion with @preamble in bibtex file

jmarten:
'@preamble' is not necessary and can be ignored like lines with '@string' or ‘@comment‘.
Additionally fixed TypeError: cannot concatenate 'str' and 'dict' objects.


lorenzhs:
I'm running into this TypeError as well, it would be great to have this merged.

msiniscalchi:
Thanks---I added "@preamble" to the ignored directives. The other issue was also resolved. Closing.

PR #327, opened 2014-02-19
Add Additional File Extension to delete_temp_files.py

acroz:
Also delete .tdo files, which are created by the latex todonotes package


ig0774:
I'm closing this as the temporary file extensions are now configurable through a setting rather than needing to change the Python code.

PR #329, opened 2014-02-21
ctrl+l (super on OSX), r adds a \ref{}

chid:


r-stein:
I think this should just be set by the user, if he wants this behavior.

PR #333, opened 2014-02-23
Improves syncing with evince

jonasmalacofilho:
* fixes forward search problems: use file system encoding and better URI quoting
 * fixes backward search problems: do not rely on URI quoting scheme
 * send column number to evince, even though this does not work on evince 3.10.0


ig0774:
Thanks for this! I realise this is quite late, but I've integrated most of these fixes into the latest release.

jonasmalacofilho:
Don't mention it, I'm happy to hear that the fixes have somehow been integrated!

PR #360, opened 2014-03-15
added documentation for root-document

Fabbi:
added documentation for setting root-document in sublime-project


ig0774:
Thanks for this. I think we've got this covered in the README now.

PR #362, opened 2014-03-17
cite completion was broken because of utf8 problems in tex file

JensLincke:
catched UnicodeDecodeError


ig0774:
Hopefully we don't come across this exception anymore... Everything is opened by default in utf-8.

PR #366, opened 2014-03-18
fix spaces in keyword for cite autocompletion

mb1248:
This patch trims the cite-keyword. ( foo) looks ugly


ig0774:
Thanks for this. It's been merged.

PR #379, opened 2014-04-07
Added setting show_panel to switch automatic open of output panel on build

davidschreiber:
`LaTeXTools.sublime-settings` now takes an extra top-level options `show_panel` (which is set to `true` in default settings file). If set to `true` the output panel will be opened on build (as it is done if the option is omitted / which is the same behavior as previously). If set to `false` the output panel will stay hidden on build.


ig0774:
Thanks for this. We've actually got something quite similar implemented now.

PR #388, opened 2014-05-06
docs for the project settings approach for TEXroot

joernhees:
this is quite a cool feature if you're working with several people and editors on some tex sources...
currently there's only a placeholder in the readme, which this should fix.


ig0774:
Thanks for this. I think we've got this covered in the README now.

PR #412, opened 2014-07-10
Fix: platform-specific `builder_settings` bug

mkanai:
Since platform-specific settings are specified under `builder_setting['osx']` in `LaTeXTools.sublime-settings`, `TraditionalBuilder` doesn't resolve custom settings.

```
"builder_settings" : {
(...)
        // Platform-specific settings:
        "osx" : {
            // See README or third-party documentation
        },
(...)
```

Just added `plat` key before `.get` method.

### DIFF
--- a/builders/traditionalBuilder.py
+++ b/builders/traditionalBuilder.py
@@ -53,9 +53,9 @@ class TraditionalBuilder(PdfBuilder):
 			default_command = DEFAULT_COMMAND_WINDOWS_MIKTEX
 		else: # osx, linux, windows/texlive, everything else really!
 			default_command = DEFAULT_COMMAND_LATEXMK
-		self.cmd = builder_settings.get("command", default_command)
+		self.cmd = builder_settings[plat].get("command", default_command)
 		# Default tex engine (pdflatex if none specified)
-		self.engine = builder_settings.get("program", "pdflatex")
+		self.engine = builder_settings[plat].get("program", "pdflatex")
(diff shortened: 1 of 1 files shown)


mkanai:
Any motions on this? Did you ever checked your `builder_settings` works as specified?

ig0774:
I agree that it could be useful to allow platform-specific command functions (especially since the default command is actually tied to platform); however, this change actually changes the documented behaviour without updating the documentation.

From the README:

"You may also have to set the `command` option in "builder_settings"..."

"**Customizing or replacing the compilation command** (latexmk or texify) is also possible by setting the `command` option under Builder Settings"

I think the way forward may be to allow the `program` and `command` settings to be overridden on a per-platform basis, but to also allow them to also be set immediately under `builder_settings` if the `command` or `program` is not platform-specific as it works today. I'd imagine this working in a similar fashion to the inheritance behaviour of #609.

mkanai:
> ...is also possible by setting the `command` option under Builder Settings

Sorry for my misunderstanding of the documentation. Okay, **under `builder_settings`** is literally *under* the `builder_settings`. As you mentioned, since `traditionalBuilder.py` changes build command according to platform, I thought `bulder_settings` should be platform-specific. 

if this implementation is intentional, **+1 for overridable settings**. I think it's totally natural to tie command with a platform (especially for those who use LaTeXTools in multi platforms.)

PR #420, opened 2014-08-19
Koma script sections

ig0774:
This is just a simple change to the Symbol List - Sections file to add support for \addchap, \addsec, \addsubsec commands from KOMA-Script to the section list. Doesn't change support for \chapter, \section, \subsection, etc.


ig0774:
On reflection, this actually doesn't work without changing either changing the scope or (even worse) the `LaTeX.tmLanguage` file.

Issue #283, opened 2013-11-04
Lualatex and fontspec


fmetzger:
If a document with the %%!TEX program = lualatex directive is built and the fontspec package is present, fontspec will throw a "cannot-use-pdftex" error (The ST output will only show the first line of the error, which is just "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!").

This could be fixed by passing latexmk the  "-pdflatex=lualatex" parameter.

PR #421, opened 2014-08-25
Lenient Comment Parsing

ig0774:
These changes make the parsing of `%!TEX program` (for the traditional_builder) and `%!TEX root` more lenient. In particular:

* These comments will be handled as long as they are in the first 20 lines of input (consistent with TeXShop)
* Initial whitespace (only) before the comments is allowed
* Multiple comment marks are allowed, e.g. `%% !TEX program = ...`, closing #283


xacid:
Hi, I just wrote a patch that is almost identical to your "Initial whitespace (only) before the comments is allowed". My patch is very simple, and you can check it here:
https://github.com/xacid/LaTeXTools/commit/904b1bae5203fa69cc532b9648e56ce0e627094f
I think your patch is very useful.

ig0774:
This is out of date. May look into doing something similar but as a clean patch.

PR #423, opened 2014-08-27
Better cite-completions

ig0774:
These are the same changes from #422, but done with a clean history so that it can be pulled without affecting anything other than citation completions.


ig0774:
Last four commits make formatting of displayed fields a bit more consistent and address #496.

aleth:
Nice patch queue! You might also like to add something like #315.

ig0774:
@aleth: Thanks! I think my preference for doing something like that, however, would be to just expose all the fields to the format panel, which shouldn't be too hard against these patches. I might have a look at doing that, as it seems useful.

ig0774:
Happier with the plugin version of this, so I'll kill this pull request.

Issue #258, opened 2013-08-25
Feature suggestion: reading .bib file under ~/Library/texmf/bibtex/bib/ in OS X (and corresponding locations in other OS)


ntrolls:
I keep a single bibtex file, which are accessed by all my TeX documents from ~/Library/texmf/bibtex/bib/ directory in OS X (well, actually, the bibtex file is symlinked from my DropBox into that directory for inter-machine synchronisation, but that is a separate matter). I find the citation completion system really indispensable, but it only looks for the .bib file from the current directory. It would be fantastic if the plugin could also look at the shared location.

I noticed that you will soon allow full customisation of the build process, so this may be doable through that, I guess. Just wanted to leave this on record.

ig0774:
To be clear, this should be implemented since #425 was merged.

Issue #60, opened 2012-08-10
Cannot auto-complete citations using a bib file in the texmf tree


tpoisot:
Hi,

I have a file called library.bib in my local texf tree, so that I can use is in all my documents. When I try to use `cite_` in LaTeX tools, it fails because it looks for the file in the local directory, which obviously fails.

Having bib files in the texmf tree is standard practice in LaTeX, any way the plugin can conform to it?

ig0774:
This was fixed quite a while ago now.

PR #425, opened 2014-08-27
Bibfiles in TEXMF Tree

ig0774:
Finding bib files in the TEXMF tree seems to come up frequently---e.g. #314, #60 and #258. This seemed like a simple, clean way of implementing it: just execute `kpsewhich` with a file name and let it do the search. In the interest of performance and minimising damage to already existing projects, I've set it so that `kpsewhich` only gets run if LaTeXTools cannot otherwise find a specified bib file. If `kpsewhich` cannot be found on the path, it fails silently. I've tested this successfully on both ST2 and ST3.

### DIFF
--- a/kpsewhich.py
+++ b/kpsewhich.py
@@ -1,0 +1,15 @@
+from subprocess import Popen, PIPE
+
+def kpsewhich(filename, file_format=None):
+    try:
+        if file_format is None or type(file_format) != type(''):
+            p = Popen(['kpsewhich', filename], stdout=PIPE, stdin=PIPE)
+        else:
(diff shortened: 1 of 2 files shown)


msiniscalchi:
This is a great contribution, thanks! Please tell me whom I should thank in the README---I'm writing "ig0774" for now.

msiniscalchi:
Thanks again @ig0774 --- see the above message. People in #314, #60 and #258 , if this solves your issue, close it.

ig0774:
@msiniscalchi Credit should be Ian Bacher. Thanks for the great plugin!

PR #431, opened 2014-09-19
Add LaTeX-cwl auto-completion support

btstream:
* "TEXroot" option in project file can use relative path to project's root path. This feature ONLY works for ST3!

* Add LaTeX-cwl auto-completion support.  Works for both ST2 and ST3.

By default LaTeX-cwl auto-completion will be turned on automatically when LaTeX-cwl
package is installed.

You can also add configurations as follows in configuration files:
```json
"cwl_list" : ["tex.cwl", "latex-209.cwl", "latex-document.cwl",
"latex-l2tabu.cwl", "latex-mathsymbols.cwl"]
```
which means the cwl list used for LaTeX-cwl auto-completions.


msiniscalchi:
I'm going to merge @ig0774 's related patch.

PR #432, opened 2014-09-19
Add Fill all feature

btstream:
Add Fill All Features. This feature is similar to latexing’s fill all
command.

NOTE: Default short cut is set to ```Super+l, Super+f```

It  support \input, \include, \bibliography, \bibliographystyle,
\documentclass, \usepackage, \includegraphics, \addbibresource, \cite
and \ref.

There also has a latex_gen_pkg_cache command， which used to generate
cache file used in fill all command for installed latex packages,
bibtex style templates and document classes.


Kreger51:
This is awesome! Only missing feature IMO.

If I may suggest a slight improvement, I would turn file types to lowercase when checking them against ``image_types``. The main reason is that snipping tool under Windows saves snips in ``PNG`` (yup, all caps) by default. 

Cheers!

msiniscalchi:
Sounds impressive! Thanks!​

On Sat, Mar 28, 2015 at 4:11 PM, Selim Belhaouane <notifications@github.com>
wrote:

> This is awesome! Only missing feature IMO.
>
> If I may suggest a slight improvement, I would turn file types to
> lowercase when checking them against image_types. The main reason is that
> snipping tool under Windows saves snips in PNG (yup, all caps) by
> default.
>
> Cheers!
>
> —
> <https://github.com/SublimeText/LaTeXTools/pull/432#issuecomment-87309210>
> .
>



-- 
Marciano Siniscalchi

Kreger51:
Hmm. Just noticed path separators under Windows are currently "\". While it is the Windows way, it is unfortunately not the LaTeX way. I know there exists a ``posixpath`` module with basically the same functions as ``os.path``. I'll try and look into it this weekend and see if I can save you the trouble. Won't be able to test under UNIX/OSX though.

msiniscalchi:
OK, I should be able to pull this shortly. However, I'd like to ask for your help:

1: the keymap files still include folding shortcuts; I'd be happier if those could be removed from this commit.

2: with reference to @Kreger51 's comment, is this working on Linux / OS X, or is it Windows only so far? 

Thanks @btstream !!!

btstream:
I have already removed the shortcuts of folding command. 

"image_types" can be set in LaTeXtools.sublime-settings, such as:

```
image_types: [png,PNG]
```
then lowercase or uppercase will be all checked.

About Windows, I usually do not work on it. I'll do some other further test on it.

btstream:
Separators on Windows are now working. I do not use  ```posixpath``` module, I just replace the '\' separators when the OS is Windows. :)

Kreger51:
Cool!

msiniscalchi:
@btstream so I pulled this into a branch called fillall-fold. I had to pull in your folding commit (because we need get_region). 

I did some more testing. It turns out that things work under Windows; however, I had trouble with \usepackage{} completion under OS X Yosemite. I'll check again tomorrow.

Also, I played with this feature a bit. I like completion for images, packages, classes, and tex files; however, what's the advantage of using c-l,c-f for citations, when there is a more powerful command (c-l,x) already available---and with autotrigger to boot?
 
My inclination would be to remove "fill-all" for citations, to avoid confusion.

btstream:
I have removed the get_region dependence from folding.

About the fill with \cite and \ref, I have changed them to using 'latex_cite_ref' command. I think to keep a unified interface or shortcut for all fill action would be a better experience for users :). 

And the \usepackage{} is working fine on my Yosemite. Is ```kpsewhich``` in the ```PATH``` environment variable?  When launching ST, environment variables are different from systems' environment variable, I think that would be the problem.

btstream:
I have change the way to search `kpsewhich`. Now if the setting of `texpath` contains the path containing `kpsewhich`, it will work.

PR #445, opened 2014-10-21
fix jumpToPDF.py, replace the absolute path to displayline

yipeipei:
Just call `displayline ` and let symbolic link do the rest work.


msiniscalchi:
THis is taken care of in another pull request, which I just merged. Thanks though.

PR #448, opened 2014-10-26
Extend 'makePDF.py' to also accept the file extension '.tikz'. It is now...

PeterPablo:
... easy to extend the list of accepted file extensions by simply extending the string.

The motivation for this is that I also want to compile TikZ code that is generated from "matlab2tikz". Those files need to have the extension ".tikz" because I also include them in my main latex file using the CTAN package "tikzscale".

This is my first pull request. Suggestions how I should improve the process are highly welcome!


PeterPablo:
I added #489 and #490 to split the commits up. Would you want a setting for `-shell-escape`? Please comment in #490.

PR #449, opened 2014-10-27
Add biber support for simpleBuilder

raimund-schluessler:
Might solve #10


Doggie52:
I would love this too! Please can this be looked at!

lockie:
:thumbsup: +1

basvandenbrink:
+1, would be nice if this would be merged.

Andi-Lo:
Would also love this!

msiniscalchi:
Folks, the simple builder is supposed to be just that---a simple builder :) That is, just example code to get builder writers started. So, simplicity is by design. That said:

1. If the objective is  to avoid the traditional builder, there is a patch pending from @ig0774 , namely #617 , which implements a "script"-based builder. This may be a way to address the issues you are having, and it's something that we are going to merge eventually.

2. If you'd like to take the simple builder and add features, that would actually be great. We would definitely consider patches. However, please do so by creating a *new* builder. I would like to keep the simple builder, well, simple. If you do decide to code a new builder, check the README and my blog post on the matter at tekonomist.wordpress.com ; builders can live in User, so they are easy to develop. 

I hope you are OK with the above. Please close.

PR #466, opened 2015-01-08
Update latex_cite_completions.py

thejakeyboy:
Line 287 was failing when trying to do cite completions. The reason:
     TypeError: Can't convert 'dict' object to str implicitly


thejakeyboy:
I noticed that a print statement was failing when I tried to do cite completions.  Line 287 (print ("Previous record " + entry)) failed with a typeerror. I suggest you take out the print statement or just remove the implicit casting of dict to str.

msiniscalchi:
I'm going to print just the keyword, rather than the entire dict. Thanks for spotting this!

PR #468, opened 2015-01-21
Add support for zathura and an option to choose a PDF viewer

ssuerbier:


ig0774:
I've added support for Zathura in v3.8.0.

PR #476, opened 2015-03-06
Make from_keybinding default to false; Find Skim if not located at default path

szhu:
From [README.markdown](https://github.com/SublimeText/LaTeXTools/blob/master/README.markdown)

> If you have installed Skim in a non-standard location, there is not much you can do short of hacking the `jumpToPDF.py` file (**not supported!**). This will change in the near future. 

I couldn't wait for it, so I did it myself. Let me know if it looks good!

### DIFF
--- a/jumpToPDF.py
+++ b/jumpToPDF.py
@@ -62,8 +62,13 @@ class jump_to_pdfCommand(sublime_plugin.TextCommand):
 		if plat == 'darwin':
 			options = ["-r","-g"] if keep_focus else ["-r"]		
 			if forward_sync:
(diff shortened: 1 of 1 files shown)


msiniscalchi:
I would like to merge this. However, why is the from_keybinding patch necessary? It shouldn't be. Did you run into errors?

szhu:
Like, it was erroring if `from_keybinding` wasn't given.

I'm not an expert at this, and neither are most users, so I figured giving it a default value couldn't hurt.

szhu:
Thanks! Sorry I was editing my previous comment when you merged :)

PR #480, opened 2015-03-18
Pretty error log for MikTeX by enabling batchmode

williamledoux:


msiniscalchi:
Sorry, but log file parsing is brittle as it is; I'm afraid that changing to batchmode will have unforeseen consequences. Closing for now.

PR #485, opened 2015-04-09
TEXroot relative to project settings

### LINKED ISSUE #152: Bad cwd on project open
When Sublime loads the last project on startup, getTeXRoot.py returns an incorrect path for the main file. If I put "TEXroot": "main.tex" into the .sublime-project file, then getTeXRoot.py returns something like c:\windows\system32\main.tex, which is obviously wrong.

Suggested fix:
replace
root = os.path.abspath(view.settings().get('TEXroot'))
with
root = os.path.join(re.sub(r"\\[^\\]+$", "", view.file_name()), view.settings().get('TEXroot'))

wronex:
I've modified `get_tex_root()` to take the project-file's location into account when searching for the master file for projects that define a `"TEXroot"` setting. This solves two issues. Firstly, it is now possible to compile multi-file projects from included tex files that are located in subdirectories. Secondly, it is now possible to open multiple projects simultaneously (in different windows) and compile them. Issue #152 is related.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -139,7 +139,12 @@ The default ST Build command takes care of the following:
 
 **Multi-file documents** are supported as follows. If the first line in the current file consists of the text `%!TEX root = <master file name>`, then tex & friends are invoked on the specified master file, instead of the current one. Note: the only file that gets saved automatically is the current one. Also, the master file name **must** have a `.tex` extension, or it won't be recognized.
 
-There is also support for project files; this is to be documented.
+There is also support for project files. A master file can be specified in the project's settings section by setting the `"TEXroot"` key. In the following example `"main.tex"` is the project's master file:
+
+	"settings":
+	{
+		"TEXroot": "main.tex"
+	}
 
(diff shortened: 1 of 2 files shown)


msiniscalchi:
I would like to merge this, but I don't fully understand how it works. In particular, would the code for reading the "global" settings ever be reached? Are "global" settings the ones from LaTeXTools.sublime-settings or other settings files? If so, I don't see how it would make sense to specify the TeX root in a global settings file...

I'm sorry if the above is naive---I don't actually use projects myself, though I can see they may be very useful to define project-specific things such as build engines (this should be possible right now), or the  location of bib files (not yet implemented).

ig0774:
@msiniscalchi: The "global" settings are the settings specific to the view. These default to the values specified for the project in its `settings` section. Apart from that, they could also be set by invoking the API. They are definitely *not* loaded from any settings file. Basically that block is pretty much dead code unless somebody somewhere was relying on setting `TEXroot` even though they didn't have a project file, e.g., in a custom `EventListener` or something similar.

Incidentally, I don't think build engines are configurable in project files yet, at least not while making use of `makePDF.py`...

wronex:
It was merely an attempt at backwards compatibility... but I've updated the pull request and removed the dead code. The code is now ST2 safe, but, the issue still remains for ST2 users since ST2 does not support ``view.window().project_file_name()``.

In hindsight, I think that ``% !TEX root = `` comments are a better solution. Maybe you should ignore this pull request and instead remove support for settings files.

ig0774:
Since I've merged #629 which does the same thing but also works on ST2, I'm closing this.

PR #489, opened 2015-04-13
Extend 'makePDF.py' to accept fileextension 'tikz'

PeterPablo:
It is now easy to extend the list of accepted file extensions by simply extending the string.


msiniscalchi:
Duplicate of #522 , which was just merged and will be pushed in a few days.

PR #494, opened 2015-04-17
Delete Temp files with double extension (like .synctex.gz)

### LINKED ISSUE #492: Synctex files not cleaned correctly
`delete_temp_files.py` fails to delete the `.synctex.gz` files on my system because it uses `os.path.splitext(file_name)` which returns the file extension as `.gz` rather than `.synctex.gz`. One solution may be to simply check if the file name ends with `.synctex.gz` since this is the only extension in the list that has more than one period.

balu-:
this should fix #492


r-stein:
you can close this now (superseded by https://github.com/SublimeText/LaTeXTools/pull/520)

msiniscalchi:
Fixed by merging #520. Closing.

PR #500, opened 2015-04-28
Bugfix: Make kpsewhich work on OS X

### LINKED ISSUE #498: kpsewhich.py fails to call kpsewhich on OSX
Unfortunately, the new "also look in the texmf tree" feature is broken on OSX, because in kpsewhich.py, `Popen(['kpsewhich', filename], stdout=PIPE, stdin=PIPE)` fails to find kpsewhich and throws an OSError.
This is despite the same kpsewhich command line working properly when tried in a shell.

I attempted to fix this by adding "shell=True" to Popen, but when one does that, kpsewhich seems to be found but its output is not captured correctly (you always get an empty string).

Hardcoding the absolute path of kpsewhich in kpsewhich.py does work, but is an ugly workaround.

ig0774:
Pull request to fix #498. Will add some details there.

### DIFF
--- a/kpsewhich.py
+++ b/kpsewhich.py
@@ -55,6 +55,8 @@ def kpsewhich(filename, file_format=None):
         path = p.communicate()[0].decode('utf-8').rstrip()
         if p.returncode == 0:
             return path
+        else:
+            sublime.error_message('An error occurred while trying to run kpsewhich. TEXMF tree could not be accessed.')
     except OSError:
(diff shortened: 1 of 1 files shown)


aleth:
This really needs to be merged...

ig0774:
@msiniscalchi: Can this get pulled? It fixes the bug with kpsewhich from the last release.

ig0774:
Mistakenly closed this... It still need to be merged.

Issue #496, opened 2015-04-24
Cite panel wildcards support for `date`, `journaltitle` in biblatex?


dofine:
I'm using Zotero generated biblatex .bib file, which looks like

```latex
@article{brumer_control_1986,
	title = {Control of unimolecular reactions using coherent light},
	volume = {126},
	issn = {0009-2614},
	url = {http://www.sciencedirect.com/science/article/pii/S0009261486801713},
	doi = {10.1016/S0009-2614(86)80171-3},
	abstract = {A method of controlling product ratios in unimolecular reactions using the coherence of lasers is presented. Theory, computational results on diatomic and polyatomic molecules, and an outline of a proposed experiment are included.},
	pages = {541--546},
	number = {6},
	journaltitle = {Chemical Physics Letters},
	author = {Brumer, Paul and Shapiro, Moshe},
	date = {1986-05-23},
}
```

There is `date` and `journaltitle`, so I can't get the journal name or published year from cite panel. 
`{{journal}}` and `{{year}}` always show as ???? in the panel.

Maybe I'm using biblatex the wrong way?

ig0774:
These changes have been made in the latest version.

PR #502, opened 2015-04-30
Support date and journal title

ig0774:
Quick fix for #496 using the vanilla cite completions. Will populate the `journal` key with the `journaltitle` field if the `journal` field is empty and similarly populate the `year` key with the year extracted from the `date` field if the `year` field is not set.


ig0774:
These changes were incorporated into #519.

geniiiii:
This doesn't seem to work for me. I'm refering to items in exported biblatex .bib files from Zotero and always see ???? instead of the year.

PR #504, opened 2015-05-08
Add LaTeX-cwl auto-completion support

ig0774:
Not trying to steal credit for others work. This is @btstream's pull request in #431. However, #431 seems to get a little clobbered with the other features @btstream was proposing, so I've created a clean history that removes dependency on anything except what's already in `master`.

Two notes on this, though: first I've refactored the `get_Region` function that all of @btstream's changes seem to use into it's own module. Second, if the changes for #432 are merged, it will be necessary to modify `latex_cwl_completions.py` to import `TEX_INPUT_FILE_REGEX` from `latex_input_completions` and add that to `ENV_DONOT_AUTO_COM`.

Finally, although in theory this should add `\` to the list of auto-complete triggers for LaTeX files, it's been more reliable on my machine to just go ahead and add this entry to the `auto_complete_triggers` setting in my user preferences file:

    {"selector": "text.tex.latex", "characters": "\\"}


ig0774:
Ok, I did actually find a serious improvement to make: on ST2 this wasn't properly working if the cwl files contained utf-8 characters not in the ascii range. This is fixed in commit 67f954a.

r-stein:
This is very nice and helpful, just two points:

- commands without args insert a `\` before, if it is missing
  commands with args does
cwl_compl_cmd_example
- It would be nice to add a special environment support by parsing the cwl files and changing the completion if the predecending text ist `\begin{` or `\end{`.

ig0774:
@r-stein: Thanks for pointing out the issue with `\`. I hadn't noticed, as I've just been using the `\` as an autocomplete trigger. It turned out to be fairly complex to support both mechanisms, but I *think* I've ironed it out in the last three commits to this branch (at least it works for me...).

Re: special support for environments: that's a great idea! I've also thought about just doing something to auto-insert `\end{..}` when filling in `\begin{..}`. I'll stick it in the queue to look into when I have enough time.

msiniscalchi:
Ian, is it safe to merge this code at this time? Thanks!

On Sat, Nov 14, 2015 at 6:39 AM, ig0774 <notifications@github.com> wrote:

> @r-stein <https://github.com/r-stein>: Thanks for pointing out the issue
> with \. I hadn't noticed, as I've just been using the \ as an
> autocomplete trigger. It turned out to be fairly complex to support both
> mechanisms, but I *think* I've ironed it out in the last three commits to
> this branch (at least it works for me...).
>
> Re: special support for environments: that's a great idea! I've also
> thought about just doing something to auto-insert \end{..} when filling
> in \begin{..}. I'll stick it in the queue to look into when I have enough
> time.
>
> —
> <https://github.com/SublimeText/LaTeXTools/pull/504#issuecomment-156693290>
> .
>



-- 
Marciano Siniscalchi

ig0774:
With the last commit (which fixes a couple of errors), it works (I've tested on ST2 / ST3 on Mac and ST3 on Windows).

r-stein:
For some weird reasons ST removes all other completions, if you show `\` or `{` to the user. (first value of the completions tuple) Hence with this addon I only see latex commands and not my already used words. This is my desired behavior, when I am writing a command, but not in normal text.

I think a first approach would be to check, that the completion is precedented by `\`, e.g. with:
``` python
bpoint = point - len(prefix)
char_before = view.substr(sublime.Region(bpoint - 1, bpoint))
if char_before != "\\":
    return []
```

In my opinion it can be pulled after applying this change and a future pull request can optionally add the completions always by examining ST's behavior.

PR #510, opened 2015-05-27
Update readme.markdown to a better installation manual. 

Limro:
- Users can now clearly see step by step what they need to do.
- All programs now have a link to where to download the software.


ig0774:
I think this is handled by the latest version of the documentation. Thanks!

PR #514, opened 2015-06-08
Update plugin version

margaritis:
The current version is misleading and gives the impression that you are actually installing an older version of the plugin.


msiniscalchi:
I actually fixed this. Closing.

PR #491, opened 2015-04-16
Ignore .git and .svn folders to prevent corrupting their contents

RomeshA:
`delete_temp_files.py` deletes files with matching extensions descending recursively into folders **including .git**. On my machine, cleaning the LaTeX temporary files results in corruption of the associated git repository because git includes `.idx` files.  I've modified `delete_temp_files.py` to ignore folders whose names match a list of folder names, currently including `.git` and `.svn` by default which prevents inadvertent damage to the repositories.

PR #516, opened 2015-06-10
more temporary filetypes to be deleted

mcnesium:
added some more temporary filetypes


ig0774:
I wonder if the list of extensions hasn't gotten to the point where this would be better implemented through a user-customisable list in the settings, especially as we increasingly run the risk of having a default list that clobbers non-LaTeX-generated files depending on the user's setup (as happened with #491)

msiniscalchi:
Yes, that's what I'd like to implement.

ig0774:
This can be closed since #520 basically covers this

mcnesium:
kthx

PR #506, opened 2015-05-14
Extended cite command matching

ig0774:
I separated this out from #423 so that that change was more sane and because the regexes used in this branch are... horrendous. But they should work with or without #423. Basically, this adds support for all the various cite commands from biblatex, natbib, csquotes, and apacite. As such it should resolve #103, #178, #464 and #392.

PR #519, opened 2015-06-12
Better cite-completions (plugin version)

ig0774:
This merges the changes in #423 and #422 into a separate plugin using the plugin system from #618. This should enable users to  provide their own bibliography plugins to override any of our decisions.

**Bibliography Features:**
 * More robust parsing support for *.bib files
 * UTF-8 support (should be fairly complete)
 * Translates LaTeX special characters into corresponding UTF-8 for UI displays, e.g. `{\'e}` will display as `é`.
 * Some support for loading fields from crossref'ed entries

In this version, `latex_cite_completions` provides space for two user-implementable methods: `get_entries` and `on_insert_citation`. The default implementations for these are in `bibliography_plugins/traditionalBibliography.py` (if it's easier to follow [here](https://github.com/SublimeText/LaTeXTools/compare/master...ig0774:bibliography_plugins) is a diff between a version of this plugin system and the current master that uses the same parsing code as the current shipping `latex_cite_completions`).

For now, users can use this to implement a bibliography plugin using this skeleton:

```python
from latextools_plugin import LaTeXToolsPlugin

class MyBibPlugin(LaTeXToolsPlugin):
    def get_entries(self, *bib_files):
        # do whatever processing to return sequence of bibliographic entries
        # returned entries should be dict-like objects that map between an entry key and its value
        # for backwards compatibility, the citekey should be mapped to the `keyword` entry
        pass

    def on_insert_citation(self, citekey):
        # do whatever processing you want with the citekey
        pass
```

And then adding an entry `bibliography_plugins` to their `LaTeXTools.sublime-settings` file with the value `['my_bib', 'traditional_bibliography']`. It is recommended to keep `traditional_bibliography` at the end of this list as a fall-back implementation. Additionally, the plugin itself will need to be in a known location. The easiest method to do this is to save it in a file with the `.latextools-plugin` extension in the User packages directory. All files with that extension will be automatically found and loaded.

Fuller details on bibliography plugins are available in `latex_cite_completions`. Fuller details on the plugin system overall can be found in `latextools_plugin`.

I have some plans to look into expanding bibliography plugins to allow them to replace the `find_bib_files` functionality and, using `on_query_context`, hopefully allow them to replace the regex system for detecting if we are in a citation command (since, as can be seen from #506, this can get really, really messy).


msiniscalchi:
This is fantastic!!! I do need to take the time and implement all the wonderful additions you have contributed. I'm travelling now, but hopefully in a couple of weeks things will slow down enough for me to catch up. Thank you!

Date: Fri, 12 Jun 2015 01:01:36 -0700

ig0774:
This pull be shortened by moving from `pybtex` back to a custom bibtex parser I've developed. The comparison between this and the current master can be found [here](https://github.com/SublimeText/LaTeXTools/compare/master...ig0774:better_bibliography_find_files_custom_parser_plugin).

ig0774:
99be2b2 changes this from using pybtex to a custom bibtex parser I've written which (hopefully) should provide the same level of functionality (I've been using it for four months so far myself).

My rationale for making this change is:
 * It's simpler than pybtex and we'd only ever leverage a fraction of pybtex's functionality
 * Support pybtex on both ST2 and ST3 required hacking some of the pybtex code because of the change in Python version (they use 2to3 to support both Python 2.x and Python 3.x; at the time I started this, PackageControl didn't have the dependencies feature and I'm still not fully sold on the idea that that's the way to go).

geniiiii:
The year title extraction (biblatex files) doesn't seem to work for me. I'm refering to items in exported biblatex .bib files from Zotero and always see ???? instead of the year.

r-stein:
@geniiiii Please create a new issue and include a minimal example file. (either as code block change the extension from bib to txt and upload it.)

PR #520, opened 2015-06-12
User configurable temporary file extensions

ig0774:
Adds two settings to LaTeXTools settings, `temp_files_exts` and `temp_files_ignored_folders` which are configurable either in the global user-settings file or on a per-project basis. Basically this should prevent the need to create a pull-request just to add file extensions to be deleted.

I've also incorporated the fixes from #494 and #491, as both of those are bugs, and made a couple of small improvements to the code.


msiniscalchi:
Merged, thanks!

PR #521, opened 2015-06-12
Fill-all support

ig0774:
Originally this branch was just for my own purposes and the majority of the work is @btstream's fantastic Fill All feature. However, this version adds some small features and fixes a few ST2-related bugs.

**Features:**
  * If you run the Fill All command with the cursor in a \cite{} or \ref{}, it will blank out the current value and run the proper automagic fill. This could probably be improved, but it seemed more useful to do something in this situation than nothing (latex_ref_cite would get run, but think it had content, so the quick-panel wouldn't display).
  * Fill All now supports bracket-based autocompletion in the same manner as we support completions for \cite{} or \ref{}. This is not dissimilar to [2ae8533](https://github.com/r-stein/LaTeXTools/commit/2ae8533b7f49909e41a7e8c61e32fb9b440a7316) from #556, but supports the whole range of the commands supported by Fill All.
  * The generated package-list will be alphabetised (instead of relatively random) and contain only unique entries (since occasionally `.cls` and `.sty` files might have the same name).
  * Supports utf-8 paths if they (somehow) make it into the path traversal (they shouldn't---TeX doesn't support utf-8 paths, but the original version did sometimes do this on a utf-8 system.
  * On ST3, we move the cache generation to its own thread so it doesn't interfere with the user. This isn't supportable on ST2 (on ST2, you can't to anything with the view outside of the main thread---technically this is probably true on ST3, but the API functions seem defer themselves to run on the main-thread), so we gracefully fall-back to running on the main thread.
  * On ST2, put the pkg_cache in the User package folder. Maybe not the nicest solution, but at least it doesn't get clobbered by package updates.
  * `image_types` can now be configured on a per-project basis

**Bugfixes:**
  * ST2 did not like the keymap files
  * `kpsewhich --show-path` has `.` as the first entry of the returned `PATH`. Makes sense for TeX, but not here. This didn't seem to cause an issue on ST3, but on ST2 `.` resolved to the Sublime application folder and we seemed to get stuck in an almost endless loop while recursively walking the directory.
  * Supports options for the `\addbibresource` command
  * More sane handling of unsaved buffers (i.e., where the TeX root is meaningless)


msiniscalchi:
I pulled it into the 'fillall' branch---I need to make sure it works under Yosemite and El Capitan (I had to set STDERR to STDOUT when calling kpsewich). Then I'll pull it into master.

Thanks a lot!!!

r-stein:
Hello, I am not sure whether I should just comment here or create an issue, but I have a few points about this new feature (which is very nice and useful):
- As written in https://github.com/SublimeText/LaTeXTools/issues/564 backslashes should be replaced by slashes on windows
- \input should also search for *.tikz files
- I think it would also be nice to add the entries into the autocompletion popup (this does not seem to be hard with the existing structure)
- I think it should enter the closing } when canceled, if it pops up by pressing { and `auto_match_enabled` is `true`

ig0774:
@r-stein: Those are good points, especially the backslash thing. I tested on Windows, but I'm not sure if I tested with subfolders. Autocompletions should be easy to add and would be consistent with the rest of the plugin. 

Re: *.tikz files, my preferred solution draws on #522 (I have a branch already, [here](https://github.com/ig0774/LaTeXTools/tree/fillall_configurable_extensions), that merges them, but without one or the other of those changes in place I didn't want to create a PR). Maybe this isn't the right idea, though?

Re: the closing `}`, the behaviour right now is the same as for ref / cite completions, though arguable they aren't doing the right thing either.

Issue #259, opened 2013-08-26
cite* errors for .Rnw files


tgouhier:
Although the new citation system for bibtex references works great on LaTeX files with .tex extensions, it fails when a file is saved with a .Rnw extension and contains a mixture of LaTeX and Sweave/knitr R code. The error appears to stem from the fact that LaTeXTools expects the file to have a .tex extension and thus fails when the file has a .Rnw extension.

PS: Commenting out lines 42-43 in file latex_cite_completions.py solves the problem.

ig0774:
Since #522 has been merged, this should now be possible by modifying the `tex_file_exts` setting.

Issue #228, opened 2013-07-10
No bib files found!??


striatum:
Hi!
I am bit of a newbie in Mac, Sublime, LaTeXTools, so, please, bear with me and help:
I am using
\documentclass[man,11pt,a4paper,apacite,floatsintext]{apa6}
and if I invoke \cite{ I am getting an automatic message that there is no bibtex file. However, it exists, and, inf fact, file compiles fine. So, I really do not get it why I am getting this annoying message?! If I use something else, like /citep{ or \citealp{, no message and all is fine. only for the most basic \cite{
My bib file is in the same directory, and I tried both
\addbibresource{data.bib}
and
\bibliography{data}

Please, help!
Many thanks,
Petar

c7h:
The Problem is due to the non-working function `get_tex_root(view)` in `getTeXRoot.py` which will return the wrong document path. You can easily fix this by defining the `TEXroot` in your settings. Use an absolute path!

msiniscalchi:
Addressed in #522 , which is already in master.

chaonan99:
Hi all, I got the same error but case is a little different for me. Aside from a `main.tex` file in `$PROJECT_ROOT` and several `.tex`s in `$PROJECT_ROOT/chapter`, my reference files are in `$PROJECT_ROOT/ref`. I found that moving them to `$PROJECT_ROOT` and set `TEXroot` to `$PROJECT_ROOT` works but I still want to hold the folder hierachy. Is there any way to do that?

zhenjieliang:
To chaonan99 "add the following TEX-root magic comment at the top of each of your included files-ie each file in $PROJECT_ROOT/chapter"

% !TEX root = ../main.tex

problem solved.

p.s. I use sublime text as my tex editor.

Koppeprojects:
I'm happy to use the suggestions when I type \cite{  thank you for that.
But for some reason I needed put an absolute path instead of the two dots in
 % !TEX root = ../main.tex
So in my case there is something like
 % !TEX root = /Users/username/Documents/project/00_main.tex

Maybe this is worth mentioning as a tip for other users to try.

Issue #105, opened 2012-11-10
Add .ltx extension


ghost:
Currently `LaTeXTools` didn't support the building of files with other that `.tex` extension:

    cv.ltx is not a TeX source file: cannot compile.

Anyway, the `.ltx` is a correct and widely used extension for LaTeX documents (`.tex` used mostly for plain TeX).

ig0774:
In case anyone else stumbles on this, we have a setting, `tex_file_exts`, which can be use to make LaTeXTools recognise `.ltx` files.

PR #522, opened 2015-06-14
User configurable file extensions

ig0774:
A range of issues have been opened to deal the fact that the extension for TeX / LaTeX files is hard-coded as being just `.tex` in several places, most especially the build and cite / ref business. These changes move all the hard-coded references in the current `master` to use a standard method which loads the extensions from either the user's `LaTeXTools.sublime-settings` file or the project settings.

The main setting used to track this is `tex_file_exts`, which should be a list of file extensions.

**Features:**
 * Supports arbitrary file extensions as first-class citizens, i.e., will build, treat as `TEXroot`, add bibliographies from any subfiles containing the configured extension.
 * Will automatically set the syntax to LaTeX for any user-configured extensions (this can be turned off globally or on a per-project basis through the `latextools_set_syntax` setting).
 * Supports setting the main file via the `%! TEX root =` magic as long as the file contains a configured extension.

**Edge Cases:**
 * There's an edge case if one `\input`s or `\include`s a file where there are two files in the root directory sharing the same name but differing by extension, e.g., `somefile.tex` and `somefile.ltx` if added with a command like `\include{somefile}`; we will load the first file that matches in extension order and skip any others.
 * Although `is_tex_file()` compares using `.endswith()` it may not be safe to assume that these changes on their own work for anything other than file extensions.

Obviously new code like #432 / #521 would need to be adapted to work with this.

This should supersede #489 and address #105, #228 and #259.


msiniscalchi:
@ig0774 is this OK to merge?

ig0774:
I've updated this to work with the latest master and committed the changes.

PR #526, opened 2015-06-26
added support for dtx files in View PDF command

benjamin-weiss:


ig0774:
Just to pointing this out, but these changes are basically superceded by #522, albeit not by default.

PR #529, opened 2015-07-03
Better Bibliography Editing

ig0774:
This is a set of features intended to make it easier to edit BibTeX or BibLaTeX files in ST.

**Features:**
 * New `BibLaTeX` syntax. This is nothing special, it is just used by this code to differentiate determine whether it should treat a file as BibTeX or BibLaTeX. In addition, I've added a configuration option `use_biblatex` which can be set in either the user settings or at the project level that, if set to `true` (default is `false`) will automatically treat .bib files as BibLaTeX rather than BibTeX.
 * Smart completions for fields containing names. Basically, inside any BibTeX or vanilla BibLaTeX field that stores a name, we provide a list of auto-completions consisting of any name found in the current file.
 * Smart completions for crossref fields. For the `crossref` field or any of the vanilla BibLaTeX fields that can have one or more citekey values, we provide an auto-complete list of all the citekeys in the current file.
 * Snippet-completions for all BibTeX and BibLaTeX entry types.
 * Completions for all BibTeX and vanilla BibLaTeX fields when inside an entry.

Anyone who's familiar with the flexibility provided by Biber will realise that its very difficult to duplicate all of the possible ways in which BibLaTeX/Biber can be configured. But as long as you're sticking close to vanilla BibLaTeX, this should be helpful and (hopefully) stay out of your way if you're using heavy customisations.

### DIFF
--- a/BibLaTeX.tmLanguage
+++ b/BibLaTeX.tmLanguage
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>name</key>
+    <string>BibLaTeX</string>
+    <key>fileTypes</key>
+    <array>
+        <string>bib</string>
+    </array>
+    <key>patterns</key>
+    <array>
+        <dict>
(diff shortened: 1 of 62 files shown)


msiniscalchi:
@ig0774  this is a good candidate for merging, but there are conflicts with the current master.

ig0774:
@msiniscalchi: Should be all set now. I think the conflicts were just due to the default settings switch.

PR #532, opened 2015-07-15
Option for choosing PDF viewer

cbrunet:
I tested it on Linux with evince, okular, and qpdfview. It should work on Windows, if other viewer than sumatra exist. Not implemented for Mac, because different executable are called from different context, and I don't have Mac to test it.

### DIFF
--- a/jumpToPDF.py
+++ b/jumpToPDF.py
@@ -1,6 +1,7 @@
 # ST2/ST3 compat
-from __future__ import print_function 
+from __future__ import print_function
 import sublime
+import string
 if sublime.version() < '3000':
     # we are on ST2 and Python 2.X
 	_ST3 = False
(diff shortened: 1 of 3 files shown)


cbrunet:
Just rebased to be in-sync with master

jumpToPDF.py
```diff
@@ -104,57 +109,66 @@ def run(self, edit, **args):
...
+
+		if keep_focus:
+			viewercmd += prefs.get("keepfocusoptions", [])
+		if forward_sync:
+			viewercmd += prefs.get("forwardsyncoptions", [])
+
+		pdffile_in_params = False
+		for i, arg in enumerate(viewercmd):
+			# Replace template strings
+			if "${pdffile}" in arg:
```
r-stein: could also named `$pdffile`

ig0774:
I'm closing this as I believe its been superceded by #639.

PR #544, opened 2015-08-14
Added multifile support for package 'subfile'

erik-werner:
The 'subfile' package is very similar to 'input' and 'include', but cross-referencing between files included via 'subfile' was not supported. It was very simple to add it.


r-stein:
Thanks, this has been included.

Issue #497, opened 2015-04-24
Image editing feature request


renatyv:
My TeX documents contain multiple images (\includegraphics[width=0.8\linewidth]{images/non-linear-tfs.pdf}).
It would be nice to have the following features
1)image file name autocompletion
2)open-image-in-editor feature (from PDF viewer or sublime)

Current workflow is as follows:
1)Jump from PDF file to the source file by cmd+click
2)remember the image file name(path)
3)open the Image editor(Adobe Illustrator)
4)Open the file in the image editor

ig0774:
Assuming you're on ST3, you may be able to reduce some of your workflow via the [SideBarEnhancements](https://github.com/titoBouzout/SideBarEnhancements), since you can customise the menu to launch Illustrator with the file selected in ST. You'd obviously still need to be in Sublime and actually open a folder (or project) rather than just a file for this to work, but maybe it saves  you a couple of steps...

r-stein:
This has been implemented and can be closed.
The open image program can be defined in the `open_image_command` as described in the readme and default settings (you might want to remigrate your settings). The key binding is `ctrl-l ctrl-o` while being on the `\includegraphics`-command.

PR #556, opened 2015-09-12
Jumpto included tex files.

r-stein:
This Pull Request consists of ~~two~~ *one* feature:
- *this feature has already been implemented in https://github.com/SublimeText/LaTeXTools/pull/521 and https://github.com/SublimeText/LaTeXTools/pull/570*
  ~~autocomplete included files~~

  ~~for included files and graphics it now has a similar behavior to the insertion of bibtex.
  It will show a menu of possible files and also adds autocompletion entries.~~

- jumpto tex files

  If you are in a line, which includes a tex file(`\include{a/b/file.tex`), pressing `ctrl+l ctrl+o` will open the file in Sublime Text.
  If the file does not exists it can even create the missing folders and the file, and insert the tex root into the file(this is optionally, but by default).

Here a demo of the new features:
lt_include_tex

*The completion for images is related to issue https://github.com/SublimeText/LaTeXTools/issues/497 (I am also looking forward to extend the jumpto feature to images in a reasonable way).*


skychan:
Very cool feature, but in WIN, the slash(/) becomes backslash(\), it will autocomplete like:
    
    \input{root\foo.tex}

that leads to error.

r-stein:
I addressed the Issue with the backslash, because I also use windows (the demo is on WIN). To solve it I just replace all backslashes by slashes in the file name. Hopefully this will not create an other issue.

r-stein:
The file autocompletion seems to be covered by https://github.com/SublimeText/LaTeXTools/pull/521, but the "jumpto file" might still be useful. I personally think it can increase the productivity not only to navigate, but especially to create new files:
- Without this you have to 
  1. create the folder structure,
  2. and the file, 
  3. insert the TeX root,
  4. and then \input the file into LaTeX.
- With this feature you just write the \input command and press `ctrl+l ctrl+o`

If you want this feature I can just remove the file completion part from this pull request or create a new one with only the jumpto feature.

ig0774:
I really like the "jumpto file" feature. It's a real productivity boost. My only qualm is that it assumes that inputs are given as relative paths lower in the directory tree than the root file. It is possible to to use either absolute paths or paths relative to the root file but further up the directory tree, either of which the jumpto command will not open (harmless but annoying) or (more destructively) attempt to create the file incorrectly. Admittedly, this is likely an extremely uncommon case, but it should be possible to accommodate. I've taken [a stab at here](https://github.com/ig0774/LaTeXTools/blob/my_master/jumpto_tex_file.py), though I'm not altogether  comfortable with the hack to determine whether the root file should be treated as an absolute or relative path.  

On the image front (which would probably be quite useful), perhaps the "reasonable" thing to do is just to open the file in the default editor? This would require that the default editor for the platform is configured in a sensible way, but it could be easily implemented by running either [`open`](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/open.1.html) (on OS X), [`xdg-open`](http://linux.die.net/man/1/xdg-open) (on Linux), or either `cmd /c "start..."` or `powershell -NonInteractive -Command { Invoke-Item ...}` (on Windows). Obviously this isn't ideal, since the default editor might just be a viewer, but at least its something...

The other alternative, it seems, would be to provide a configuration option to set the command to run with some sort of substitution like `%F` for the file name. This could perhaps be provided with the default behaviour being what I mentioned above.

r-stein:
@ig0774 Thank you for the feedback and amendments. I did not consider the possibility to use absolute paths, but I will review your changes and take them into account to improve and change the jumpto-feature.

For the image part:
I think it would be a great to add the open commands into the `LaTeXTools.sublime-settings`-file, i.e. the command to open a file for every extension. Like in a switch-case statement the extension-match would be searched from top to down and if no extension is given, then command would be executed for every extension.
This configuration works with my current code on windows targeting this approach:
``` javascript
"osx":  {
  // ...
},
"windows": {
  "texpath" : "",
  // ...
  "open_image_command": [
    {
      // open jpeg files with paint
      "extension": ["jpg", "jpeg"],
      "command": "mspaint"
    },
    {
      // open pdf documents with sumatra
      "extension": "pdf",
      "command": "SumatraPDF"
    },
    {
      // no extension => default command
      // open all other files with the default program
      "command": "start"
    }
  ]
},
```
The fallback if no default command is entered or the config is missing at all, would be to open the file in Sublime Text. This seems to be reasonable, because
- at least ST3 is able to show `jpeg` and `png` files
- Using the very popular [SideBarEnhancements](https://packagecontrol.io/packages/SideBarEnhancements)-package you can define default programs for file types. Hence the behavior would integrate into the prior workflow without the need of changing the configuration.

ig0774:
@r-stein: That setup for images looks great! I mean, that's the ideal way for things to work.

Two things:

1. Does your code supports longer commands? On OS X, for example, the easiest way to open an image in PixelMator (popular image editing app) from an external program would be to run: `open -a PixelMator image.jpg`.

2. Would it make more sense for the default option to fallback to open the file in the default program if it isn't `jpg` or `png`? ST3 does a good job with them, but with, e.g., PDFs isn't terribly useful. I'm not sure SideBarEnhancements is much help either as it doesn't change what happens when you open a file in ST (again, thinking about a case where the file in question might not be immediately visible in the ST sidebar).

r-stein:
@ig0774 
1. Yes I agree it must support longer commands. Currently I just append the image name with a space char to the command and call popen, hence this is not an issue yet and hopefully won't be one with an other OS. In addition it could be reasonable to optionally write the file name into the args, because the program might not always accept an arbitrary order of the arguments. One might even want to write something like:
  ``` perl
"command": "start /i cmd /c \"convert $file ${file/.eps/.pdf/g} && SumatraPDF ${file/.eps/.pdf/g}\""
  ```
  But to support this does not seem to be a reasonable effort to me :) since everybody could write and call a bat/sh script for this purpose.
2. In the config of SideBarEnhancements you can set `"open_automatically" : true`. With this option SublimeText will not open the file in the editor, but will open the defined program instead. Hence it opens e.g. SumatraPDF without any additional click. In addition I would like to write the distinctions into the default options and not into the source code.

ig0774:
@r-stein: I finally had a chance to take a look at your currently pushed version of the open_images code and its really good. A couple of things, however:

Re 2. That's cool! I didn't know that!

Re 1. I agree supporting commands like that  monstrosity you suggest is well beyond a reasonable effort. I brought up the case of `open -a PixelMator image.jpg`, however, not to be annoying; its actually the OSX equivalent to `photoshop.exe image.jpg` or whatever.

The current `Popen` approach actually doesn't support this syntax, but its easy to fix using [`shlex.split()`](https://docs.python.org/2/library/shlex.html#shlex.split) to convert the command into a list and then append the file path to the end of the list a pass *that* to `Popen`. This has the added advantage that it works even if the path containing the root file or the command itself contains spaces.

The end result looks something like this:

```python
def run_command(command):
    command = shlex.split(command)
    command.append(file_path)
    subprocess.Popen(command)
```

Regarding a couple of your TODOs: 
 * `sublime.platform()` on OSX always returns "osx", **not** "darwin".
 * It would be a really good idea to support loading the supported image extensions from a user-configurable setting.The situation I'm imagining is where a user or a package loaded by the user takes advantage of the `\DeclareGraphicsExtensions` command to either (1) add new extensions or (2) redefine the extension load order. Checking for any such declarations is beyond a reasonable effort, but providing the user with a configuration option allows them to configure their local version to be aware of those changes.

r-stein:
Thanks @ig0774 I applied your changes.
The previous version worked with options for the commands on windows, but I think the functionality of Popen is OS dependent.

msiniscalchi:
Merging---thanks @r-stein and @ig0774 !

PR #570, opened 2015-10-18
Input Completions Fixes

ig0774:
This addresses a fews issues with `LatexFillInputCommand` and adds its autocompletions to the usual pop-up.

Thanks to @r-stein for pointing most of this out.

Issues addressed:
 * On Windows we were inserting backslashes instead of slashes for folder names. This would cause the autocompletions not to work for items
 * Due to an oversight, completions for the `\bibliography{}` would not work

New feature:
 * Autocompletions added to the autocompletion pop up (i.e. ctrl + space or cmd +space)

In addition, the `.tex` extension is omitted for `\input` or `\include`

### DIFF
--- a/latex_input_completions.py
+++ b/latex_input_completions.py
@@ -176,6 +176,50 @@ def add_closing_bracket(view, edit):
     view.sel().subtract(view.sel()[0])
     view.sel().add(sublime.Region(caret, caret))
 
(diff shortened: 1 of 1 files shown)


msiniscalchi:
I am merging this now.

PR #572, opened 2015-10-20
traditionalBuilder.py - Fix to settings and new features

saikon:
This proposed change contains: 
- A fix to the Traditional Builder to properly read settings defined under "builder_settings"
- Added feature to parse TeX options from the file (%!TEX option = <option>)
- Output printing to show the whole command used instead of only the first list item

Added tex-option parsing is only for windows since I do not have the knowledge on how different commands use the defined options.


msiniscalchi:
#576 (merged) is simillar enough; closing. (I understand the  full command is not printed; I don't want to scare users though, and the full command is always accessible via the ST console).

Issue #566, opened 2015-10-11
Feature request: Auto-hide console following build


stevebarratt123:
Apologies if this is posted in the wrong place, I couldn't find anywhere else. 

I routinely make this modification:

http://stackoverflow.com/questions/25441973/sublime-text-latextools-console-autohide

to makePDF.py whenever LaTeXTools is updated. The last few times I get something wrong when I try to re-edit the file, perhaps I'm missing some nuance of python syntax.. Anyway, I and I'm sure many others find this little mod very useful, particularly if one is very frequently building to check progress. Could this be added to the official release?

Either way, thanks for the fabulous package!

mckennapsean:
:+1:

r-stein:
As mentioned this is adressed by pull request https://github.com/SublimeText/LaTeXTools/pull/574.

Until the pull request is merged you might want to add this after line 177: (The stackoverflow answer is missing the caller attribut)
```python
self.caller.window.run_command("hide_panel", {"panel": "output.exec"})
```
In addition you could just checkout the git repository, locally commit your changes, and pull manually to avoid repeating the manipulations many times.

r-stein:
This can be closed now. (adressed in https://github.com/SublimeText/LaTeXTools/pull/574)

To enable the feature you can re-migrate your setting and then change the field `hide_build_panel` or just add the attribut to your `LaTeXTools.sublime-settings`, e.g.
``` javascript
"hide_build_panel": "no_errors",
```
to hide the build panel if no errors occur.

PR #574, opened 2015-10-22
Auto-hide the build panel

r-stein:
This pull request adds the feature inquired by issue https://github.com/SublimeText/LaTeXTools/issues/566 .
It adds an option (`hide_build_panel`) to hide the build panel after the build finished. Possible values are:

- __always__: always hide the panel even if the build failed
- __no_errors__: only hide the panel if the build was successful althrough warnings appear
- __no_warnings__: only hide the build panel if neither errors nor warnings occur
- __never__: never hide the build panel

Those options can be renamed if someone has suggestion for better and self-explanatory names.

### DIFF
--- a/makePDF.py
+++ b/makePDF.py
@@ -185,23 +185,29 @@ class CmdThread ( threading.Thread ):
 			else:
 				content.append("")
 
-			s = sublime.load_settings("LaTeXTools.sublime-settings")
-			hide_panel_level = s.get("hide_build_panel")
 			hide_panel = {
 				"always": True,
 				"no_errors": not errors,
 				"no_warnings": not errors and not warnings,
 				"never": False
-			}.get(hide_panel_level, False)
+			}.get(self.caller.hide_panel_level, False)
+
 			if hide_panel:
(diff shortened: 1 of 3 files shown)


makePDF.py
```diff
@@ -184,6 +184,24 @@ def run ( self ):
 				content.extend(warnings)
 			else:
 				content.append("")
+
+			s = sublime.load_settings("LaTeXTools.sublime-settings")
```
ig0774: ST2 raises an exception when you try to load settings [or do basically anything with the Sublime API off the main thread](http://stackoverflow.com/questions/29975674/which-operations-in-the-sublime-text-2-api-are-thread-safe). This setting should be loaded in the `run()` function of the `make_pdfCommand` class where the other settings are loaded and then accessed something like `self.caller.hide_panel_level`.

r-stein: Thanks a lot. In the future I will also test it with ST2.

msiniscalchi:
Merging this. Thanks! M

PR #576, opened 2015-10-30
Allow user to add options to traditional builder

ig0774:
This adds a feature similar to #572, i.e., parsing `%!TEX option =` in the root file, but works for both `texify` and `latexmk`.

The most obvious use-case for this is to pass `--shell-escape` to latex, which can be done by using `%!TEX option = --shell-escape` somewhere before the `\documentclass` line in the root file.

In addition, a list of command line options can be passed using the `options` builder-setting, e.g., the same effect can be achieved by having a `builder-setting` section like this:

 ```json   
"builder_settings": {
    "options": [
        "--shell-escape"
    ]
}
```

To accommodate the need to read more than one line, the traditional builder will now look for `%!TEX` directives in any line before the `\documentclass`, though it still only takes the *first* engine option encountered.

Finally, `DEFAULT_COMMAND_LATEXMK` has been changed  to properly support `--shell-escape`. I couldn't get `-shell-escape` to work with a manually-set `pdflatex`, so I've switched to using `latexmk`'s `-pdf`, `-xelatex`, and `-lualatex` switches. This shouldn't affect the user: we end up with the exact same command being run. This does, however, mean that we require `latexmk` v4.37 or later for lualatex and v4.31 or later for anything else. I don't imagine this poses a problem (v4.31 was available for TeXLive / MacTeX 2012), but its worth mentioning.

### DIFF
--- a/builders/traditionalBuilder.py
+++ b/builders/traditionalBuilder.py
@@ -15,13 +15,14 @@ import codecs
 
 DEBUG = False
 
-DEFAULT_COMMAND_LATEXMK = ["latexmk", "-cd",
-				"-e", "$pdflatex = '%E -interaction=nonstopmode -synctex=1 %S %O'",
-				"-f", "-pdf"]
+DEFAULT_COMMAND_LATEXMK = ["latexmk", "-cd", "-e", "-f", "-%E",
+					"-latexoption=\"-interaction=nonstopmode\"",
+					"-latexoption=\"-synctex=1\""]
 
-DEFAULT_COMMAND_WINDOWS_MIKTEX = ["texify", 
-					"-b", "-p", "--engine=%E",
+DEFAULT_COMMAND_WINDOWS_MIKTEX = ["texify", "-b", "-p", "--engine=%E",
 					"--tex-option=\"--synctex=1\""]
+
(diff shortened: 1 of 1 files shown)


alexcoplan:
I really need to be able to set this option now in order to use the minted package. Is there a quick fix?

msiniscalchi:
I"ll merge this request with the next update.

On Mon, Nov 16, 2015 at 5:20 PM, Alex Coplan <notifications@github.com>
wrote:

> I really need to be able to set this option now in order to use the minted
> package. Is there a quick fix?
>
> —
> <https://github.com/SublimeText/LaTeXTools/pull/576#issuecomment-157205967>
> .
>



-- 
Marciano Siniscalchi

msiniscalchi:
Merged, thanks!

Yuting-Kou:
@ig0774 Thanks for your solution, but it doesn't work for my Win10 sublime text 3+texlive setting:(

I have tried to change latextool.user-setting files and add following commands:

"builder_settings": {
 "options": [
        "--shell-escape"
    ]
}
DOESN'T WORK.

I add `%!TEX` option = --shell-escape` in the first line of my `xx.tex` file, and it DOESN'T WORK.

What else can I do to solve this problem?
>Package minted Error: You must invoke LaTeX with the -shell-escape flag. [\usepackage]

Issue #573, opened 2015-10-21
PDF preview dose not work


Ricardo-May:
I have already set my ST3 and latextools, and actually the pdf preview sometimes worked well but I have met some trouble as well sometimes it dose not jump to PDF and preview the file, what's going on with my settings? Please help me.

Code-Curious:
Same thing here, I am using ST3 on Windows and I have SumatraPDF in my Path. The output PDF does get refreshed but the it doesn't switch to SumatraPDF's window even though I configured the `keep_focus` setting to `false`

ig0774:
I'm not sure if #578 really addressed this... Can you provide any more details on what's happening? In particular, anything printed on the build panel and on the Sublime console (⌘ + ` or Ctrl + `) could be useful in working out what is going on.

Ricardo-May:
Now, this problem has not occurred anymore. I think it is already fixed for most scenarios

ig0774:
Ok. I'm going to close this for now, but please either re-open it or create a new issue if this returns.

PR #578, opened 2015-11-02
Bugfix: Jump to PDF doesn't use proper view

ig0774:
In `makePDF.py`, the final call to `jump_to_pdf` uses the active view of the current window. Since LaTeX builds can take quite a while, this is not guaranteed to be the same view that was active when the build was initiated, meaning that `jump_to_pdf` might 1) not open any file at all or 2) might open the wrong PDF file (if the view was switched to another LaTeX document).

This change saves the view used at the launch of build and uses that to call `jump_to_pdf`.

This *may* fix some of #573, but I can't really promise anything.


msiniscalchi:
Merging, thanks!

PR #580, opened 2015-11-10
Add support for bibentry's \nobibliography command

danschrage:
When searching a .tex document for .bib files to use for citation completion, look for \nobibliography command in addition to the \bibliography command. (The \nobibliography command allows one to use in-document citations without printing a references section at the end.)


msiniscalchi:
Will merge, thanks! M

PR #584, opened 2015-11-17
Update delete_temp_files.py

neok-m4700:
1. Added additionnal temp files: '.glg','.glo','.gls','.ist','.lot','.run.xml','blx.bib'
2. The line: "if file_ext in temp_exts:" won't work if dual extension e.g. .synctex.gz because os.path.splitext returns only .gz , replaced with "if any([file_name.endswith(ext) for ext in temp_exts]):"


msiniscalchi:
I just merged a related patch that achieves the same objectives.

Issue #595, opened 2015-11-23
Select Line Hotkey Changed Globally


NJBS:
The latest update has broken the default Sublime Text select line hotkey. LaTeXTools has globally overwritten the default Ctrl+l hotkey used to select a line with its double Ctrl+l style line selection normally used in Latex files. This behavior occurs across all opened files, regardless of what syntax style is being used to view them.

r-stein:
Fixed in https://github.com/SublimeText/LaTeXTools/pull/596

msiniscalchi:
Patch merged.

PR #596, opened 2015-11-23
Bugfix: LaTeXTools disables the `ctrl+l` globally

r-stein:
Added a context to the `jumpto_tex_file` command, such that is only for tex-files. Addressed in https://github.com/SublimeText/LaTeXTools/issues/595


msiniscalchi:
Thanks.

PR #601, opened 2015-11-26
Jumpto included graphics

r-stein:
This pull request adds the feature to open included images with user definable programs, as wished in https://github.com/SublimeText/LaTeXTools/issues/497. This has been discussed in https://github.com/SublimeText/LaTeXTools/pull/556.

This pull request is based in `jumpto_tex_file` and includes:
- jump to bib files: easy adaption to jump to files included with `\bibliography`, `\nobibliography` or `\addbibresource`
- the jump command will now only trigger if the selection is inside the command (having more than one commands in one line could have lead to unexpected behavior)
- opening graphics with configurable programs

Example settings:
``` javascript
// ...
"open_image_command": {
    // ...
    "windows": [
      {
        // open pdf and eps documents with inversed sumatra
        "extension": ["pdf", "eps"],
        "command": "SumatraPDF --inverse-color $file"
      },
      {
        "extension": "png",
        "command": "mspaint"
      }
    ],
    // ...
},
// ...
```
The extensions will be matches from top to bottom and can either be a list (several extensions) or a string (one extension). If `$file` is not inside the command string, the file will be appended to the command.


ig0774:
@r-stein: This is fantastic! A couple of small things for the feature for jumpto_bib_file: `\addbibresource` supports optional arguments which (if used) this will miss. Additionally, BibLaTeX has two other commands, similar to `\addbibresource` called `\addglobalbib` and `\addsectionbib` that it would be good to support. Basically, this would require changing the `_BIB_REG` to something like:

```python
_BIB_REG = re.compile(
    r"\\(?:bibliography|nobibliography|addbibresource|add(?:global|section)bib)"
    r"(?:\[[^\]]*\])?"
    r"\{(?P<file>[^}]+)\}",
    re.UNICODE
)
```
Likewise (and this makes things tricker) `\bibliography` and `\nobibliography` actually support multiple file names separated by commas. Maybe this could be dealt with separately after these changes are merged...

Nevertheless, this is all really, really useful.

ig0774:
Actually, never mind. I just realised this is actually par for what's in the cite completions right now, apart from the multiple file thing.

r-stein:
Thanks for the hint. Multiple files are now supported.

r-stein:
Which command do you recommend as the default command to open pdf files on OSX?

ig0774:
`open -a Skim.app` is probably the most sensible since we otherwise require Skim on OSX.

r-stein:
Thanks

msiniscalchi:
For this to work, if I understand correctly, the user will need to set the appropriate options in the settings file, or regenerate the settings file  If so, this should be documented in the README.

By the way, one thing I've been thinking about is a revision of the reconfigure / migrate command that can actually migrate existing settings in the user's LaTeXtools.sublime-settings. Right now only the old-style settings are migrated (and I guess by now that's no longer needed). New-style settings are simply clobbered every time the user runs the reconfigure / migrate command.

r-stein:
Now I wrote a description of the settings in the README. Without migrating the settings the image file will be opened in ST, which is useful with ST3 and png or jpg files.

In addition I agree that the migration process should be revisited. If you just rename the file `LaTeXTools.default-settings` to `LaTeXToools.sublime-settings` then the package would work with the default settings without any migration. I think this is a desirable behavior. Then one could add an entry to the default settings in the menu and optionally keep the migration feature.

msiniscalchi:
@r-stein that's essentially what the current Reconfigure/Migrate command does---it copies the default settings file to the User directory and changes the extension. I agree that the default behavior is acceptable.

r-stein:
To clarify what I said:
Just by renaming the file extension into `sublime-settings` the user would not have to migrate the settings to work with LaTeXTools.
The call `sublime.load_settings("LaTeXToools.sublime-settings")` will load both, the users and default settings and the default settings would be used only if the entry is not in the users settings (or the users settings are missing complety). In addition the developers don't need to write a default value in the code to handle the case, that the user has not migrated the (new) settings. The value would just be in the default-settings.

In addition one could add an entry into the `Main.sublime-menu` file to open the default settings.

msiniscalchi:
@r-stein aha! I see what you mean. There are two drawbacks though One  is that, by regenerating the user settings file, we also provide documentation. But this can be addressed in the README (though my preference would be to still do things nicely). 

The second is potentially more serious: the user may get the impression that it's OK to tinker with the settings file in the LaTeXTools directory---which will of course be overwritten every time there is an update. I do understand though that this is the way many ST packages work--there is a default settings file and a user settings file, and the user can load and edit either one. I think this is a usability issue. Keep in mind that many LaTeX users are not necessarily very computer-literate (I know from experience...). Sorry, I'm extremely conservative in these things.

Still, I'll keep this in mind. I'm merging this pull request BTW.

r-stein:
@msiniscalchi to address your points:

1. Yes I agree on this and think that the migrate settings function feels better than manually copying the default settings. In addition it could be extended to set some values (e.g. path to SumatraPDF) automatically or automatically open the settings file afterwards. Hence I don't argue on removing that functionality.
2. I think keeping the comment `DO NOT EDIT THIS FILE!!!` on the top of the file will prevent most users from thinking it would be a good idea to edit that file. In addition especially users, who have little experience with computers, might have problems with the "migrate your settings first"-message. 
  The settings in the user folder are not really necessary and this (small) problem could be evaded by providing the default settings. Therefore I would argue that the advantages exceed the disadvantages, even for computer-beginners.

  However the users would not be forced to read the README, which would be the major drawback in my opinion.

PS.: My intent was to make it the default settings, not to change the menu structure. Although I think opening the default settings would be a benefit. Because one could check for new settings without deleting the modifications.

msiniscalchi:
You know, I think I'm coming over to your side @r-stein :-) It may be best to do as you suggest. IIRC there is even a way to load a file in read-only mode, so that the user really has to go out of his/her way to shoot himself/herself in the foot.

Issue #603, opened 2015-11-27
Bug: Stoping compilation process does not terminate the pdflatex process


dlavoieb:
When manually stopping the compilation process which hangs due to a bad file, the `pdflatex` process keeps running in the background and never terminates.

This file is an example of file that does not compile and hangs.
[example.tex](https://files.zenhub.io/5657b1d57177c2365afba1c1)

To reproduce simply start the compilation with LatexTools and stop (press CMD+B during compile).

In terminal: `$ ps -ef | grep pdf` still show the process running and the ressource usage is quite significant (100% cpu).

ig0774:
Thanks for the report. I've make some changes to makePDF.py which should fix the issue. You can grab the modified file [here](https://raw.githubusercontent.com/ig0774/LaTeXTools/kill_child_processes/makePDF.py).

msiniscalchi:
Patch merged.

PR #604, opened 2015-11-27
Bugfix: Ensure that we kill child processes of the launched process

ig0774:
This is to fix the issue reported in #603. 

In makePDF.py we kill the current running subprocess, which is fine as long as the subprocess itself doesn't spawn any executables. Both latexmk and texify spawn executables (pdflatex, bibtex, etc.) to complete there work and if these executables hang (as in the example given with #603), the processes will not be terminated by calling `subprocess.kill()`.

On OS X and Linux, I've switched it so we use a process group and `os.killpg()` to kill the builder and its children. On Windows with its different process architecture, I've used the `taskkill` executable to kill the process tree.

In addition, I added a lock to control access to the `proc` variable because on Windows, there appeared to be a race condition where taskkill would kill `texify` or `latexmk` but the CmdThread would think that execution had completed normally.

I don't currently have access to a Linux machine, so I'd appreciate it if someone else could test that.

### DIFF
--- a/makePDF.py
+++ b/makePDF.py
@@ -244,7 +244,7 @@ class make_pdfCommand(sublime_plugin.WindowCommand):
 	def __init__(self, *args, **kwargs):
 		sublime_plugin.WindowCommand.__init__(self, *args, **kwargs)
 		self.proc = None
-		self.proc_lock = threading.Lock
+		self.proc_lock = threading.Lock()
 
 	def run(self, cmd="", file_regex="", path=""):
 		
--- a/makePDF.py
+++ b/makePDF.py
@@ -248,6 +248,7 @@ class make_pdfCommand(sublime_plugin.WindowCommand):
 			else:
 				os.killpg(self.proc.pid, signal.SIGTERM)
 			self.proc = None
+			return
 		else: # either it's the first time we run, or else we have no running processes
 			self.proc = None
 		
--- a/makePDF.py
+++ b/makePDF.py
@@ -241,7 +241,8 @@ class CmdThread ( threading.Thread ):
 
 class make_pdfCommand(sublime_plugin.WindowCommand):
 
-	def __init__(self):
+	def __init__(self, *args, **kwargs):
+		sublime_plugin.WindowCommand.__init__(self, *args, **kwargs)
 		self.proc = None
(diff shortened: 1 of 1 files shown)


msiniscalchi:
Wonderful! I'lll merge this next.

On Fri, Nov 27, 2015 at 5:42 AM, ig0774 <notifications@github.com> wrote:

> This is to fix the issue reported in #603
> <https://github.com/SublimeText/LaTeXTools/issues/603>.
>
> In makePDF.py we kill the current running subprocess, which is fine as
> long as the subprocess itself doesn't spawn any executables. Both latexmk
> and texify spawn executables (pdflatex, bibtex, etc.) to complete there
> work and if these executables hang (as in the example given with #603
> <https://github.com/SublimeText/LaTeXTools/issues/603>), the processes
> will not be terminated by calling subprocess.kill().
>
> On OS X and Linux, I've switched it so we use a process group and
> os.killpg() to kill the builder and its children. On Windows with its
> different process architecture, I've used the taskkill executable to kill
> the process tree.
>
> In addition, I added a lock to control access to the proc variable
> because on Windows, there appeared to be a race condition where taskkill
> would kill texify or latexmk but the CmdThread would think that execution
> had completed normally.
>
> I don't currently have access to a Linux machine, so I'd appreciate it if
> someone else could test that.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/SublimeText/LaTeXTools/pull/604
> Commit Summary
>
>    - Ensure that we kill child processes of the launched process
>    - Restore mistakenly deleted return
>    - Run taskkill in its own shell and hide the window
>    - Add some locking to the variable used to track the process
>
> File Changes
>
>    - *M* makePDF.py
>    <https://github.com/SublimeText/LaTeXTools/pull/604/files#diff-0> (61)
>
> Patch Links:
>
>    - https://github.com/SublimeText/LaTeXTools/pull/604.patch
>    - https://github.com/SublimeText/LaTeXTools/pull/604.diff
>
> —
> <https://github.com/SublimeText/LaTeXTools/pull/604>.
>



-- 
Marciano Siniscalchi

msiniscalchi:
Thanks!

Issue #602, opened 2015-11-26
Malformed LOG file: over/underfull


Limro:
So this popped up:
![capture](https://cloud.githubusercontent.com/assets/2076706/11425749/6ae99fca-9457-11e5-8cfc-1ba4b29b7af4.PNG)

Raw text:

```tex
	\documentclass[a4paper]{article}
	\usepackage{showframe} 

	\showhyphens{createUnspecifiedNodeErrorMarker}

	\makeatletter
	\def\code{\leavevmode\begingroup
	\let\ifcase\iftrue
	\def\or##1{%
	  \catcode`##1\active\uccode`\~`##1\uppercase{%
	    \def~{\egroup\penalty2\hbox\bgroup\string##1}}}%
	\@Alph{}%
	\@code}

	\def\@code#1{\textbf{\hbox\bgroup#1\egroup}\endgroup}
	\makeatother

	\begin{document}
	This is a longer line -- notice what happens now at the end: \code{OfThisLongPeiceOfText} 
	\code{OfThisLongPeiceOfText} 
	\code{OfThisLongPeiceOfText} 
	\code{OfThisLongPeiceOfText} 
	\code{OfThisLongPeiceOfText} 
	\end{document}
```

Need anything else?

msiniscalchi:
Patch merged.

PR #605, opened 2015-11-27
Bugfix: over/underfull box parsing

ig0774:
The current version of parseTeXLog.py assumes that the `[]` terminating the underfull box warning is on a line by itself. However, this line sometimes contains the text which caused the box badness (as in the example given in #602). This fixes the issue by assuming that the remaining text on a line that starts with `[]` or ` []` after a underfull box can be safely ignored.


msiniscalchi:
great, thanks

msiniscalchi:
Thanks.

Issue #598, opened 2015-11-23
Update broke custom completions


daftscience:
The new update seems to be interfering with my custom completion files. I'm not sure I'm doing something wrong, or what exactly is causing the issue.

Here is an example completions file.

```
{
        "scope": "text.tex.latex",
        "completions":
        [
                { "trigger": "pi",  "contents":  "\\prettyimage{width=$1\\textwidth}{graphics/$2}$0"},
        ]
}
```

Right now uninstalling LatexTools will make my configuration work. Re-installing it breaks it again.

Thanks in advance for the help

r-stein:
Do you use cwl-autocompletion?

msiniscalchi:
I haven't seen any follow-up here. Closing, but feel free to reopen if necessary.

PR #608, opened 2015-11-30
Configurable cwl autocompletion popup

r-stein:
Adds a setting to choose, when to popup the cwl autocompletion:
- __never__
- __prefixed__: if the current completion is prefixed with `\` (default)
- __always__

As already mentioned in https://github.com/SublimeText/LaTeXTools/pull/504 the completion could infer with other completions.
This PR should fix https://github.com/SublimeText/LaTeXTools/issues/598, if the attribute `"command_completion"` is set to `"prefixed"`.


msiniscalchi:
Merged, thanks!

Issue #515, opened 2015-06-08
Problem with TEXroot directive and .bib files autocomplete


odarriba:
Hello!

I'm using ST3 (I don't really know if the issue is because of this or not), latest version of LaTeXTools and latest version of Git plugin (https://github.com/kemayo/sublime-text-git).

I've found that, when using TEXroot directive in project file to set the main file to compile in the project, and and adding a Git repository (so the Git plugin recognise it):
- LaTeXTools stops recognising this directive (and compiles the file you're editing only)
- `.bib` files aren't read and shows the annoying error message every time you make a `\cite{}`.

When you delete the repository files (`.gitignore` file and `.git` folder in my case) everything works again.

I don't really know if this issue is related to LaTeXTools or Git plugin (or even ST3), but I'd like to notify it in case someone want to look at it (I'm currently writing my final project's report, so I don't have time at the moment).

Kind regards,
odarriba

ig0774:
Sounds weird... I can't reproduce that here. Do you have any project-specific settings for the git plugin? (Don't normally use that myself) What output is there on the console (ctrl + `) when you try to compile or `cite{}` an entry?

odarriba:
My project settings are this:
```
{
	"folders":
	[
		{
			"file_include_patterns":
			[
				"*.tex",
				"*.bib"
			],
			"path": "."
		}
	],
	"settings":
	{
		"TEXroot": "TFGManual.tex"
	}
}
```

And I did the following from the console
```
cd TFG/Docs/TFGManual
git init
git add .
git commit -m "Initial commit."
```

Then I open the project using ST 3 (Build 3083), compiled the main file and opened another .tex in the same project. When I wrote `\cite{` the alert message appeared anthe console shown:

```
TEX root: '/Users/odarriba/TFG/Docs/TFGManual/contents/resultados/sections/resultados-host.tex'
Searching file: '/Users/odarriba/TFG/Docs/TFGManual/contents/resultados/sections/resultados-host.tex'
Bib files found: 
[]
error: No bib files found!
```

The plugins I'm using are: 
* Git v2015.02.06.06.00.56
* LaTeXTools v2015.04.13.03.23.00

And if I delete all the repository (`rm-rf .git*`) all returns back to normal...

msiniscalchi:
Does the problem persist? If not, please close.

ig0774:
Looking back at this, I'm wondering if it isn't related to #152 and the Git stuff is a bit of a red-herring...

[...]

PR #609, opened 2015-12-01
Project Overridable Settings

ig0774:
This is mostly a back-end change to make it easier to support overriding settings at the project level. 

Basically, these changes are designed to simplify the idiom:

```python
global_settings = sublime.load_settings('LaTeXTools.sublime-settings')
setting = view.settings().get('setting', global_settings.get('setting', default))
```

To:

```python
setting = get_setting('setting', default)
```

This, of course, assumes that the setting has the same name in the view as it does in the global settings file (I've adjusted the toggles accordingly).

However, it offers two major advantages over the usual means of doing this:
 1. In cases where a missing key would raise an exception, it will return either a specified default value or `None` if no default is supplied. This should reduce a certain class of errors, assuming the underlying code can handle the `None`.
 1.  It supports overriding dictionaries of settings in an additive way. For example, I could override the Windows `texpath` setting with a project-specific `texpath` without thereby blanking out the `distro` or `sumatra` settings.


msiniscalchi:
This is excellent! Can I merge it as is?

ig0774:
I actually want to test this a bit more before merging. There seems to be some odd behaviour with reloading when this is used from sub-packages, e.g., the `latextools_utils` package.

msiniscalchi:
There were conflicts with the configurable tex extension patch. Nothing major---the import statements for both patches had to be combined. Hope all is OK!

msiniscalchi:
@ig0774 I just wanted to tell you that this feature **rocks**! I played with it a bit trying to figure out what was wrong in issue #515 and I must say it's really impressive. I haven't used project files myself so far, but that may change---esp. in conjunction with the script builder in #617 .

Issue #495, opened 2015-04-19
"Jump to PDF" loses focus (Linux, evince)


datamaan:
Everytime I build or do a "Jump to PDF" the focus goes to evince. Settings and toggles are on "keep_focus": true.

I'm running ArchLinux with KDE5 and use evince as PDF-viewer for LaTeXTools.
Is there anything I'm doing wrong?

HakanSiktas:
Same here on Windows and SumatraPDF.
> "keep_focus": true

does not work.

gutierrezps:
Same here with MiKTeX and SumatraPDF on Windows 8.1 and Sublime 2. Nor toggling through the shortcut (Ctrl+L,T,F) neither by changing the setting on LaTeXTools.sublime-settings, the focus always goes to SumatraPDF.

msiniscalchi:
This is, unfortunately, something that can't be fixed. Sumatra does not allow one to control focus via the command line, and for some reason DDE commands don't work on ST3 anymore. There's a branch on GitHub where I experimented---I tried many, many thigns, but nothing worked.
I'll document this when I get a chance. Very annoying!

Date: Sun, 14 Jun 2015 16:19:22 -0700

gutierrezps:
Thanks for the quick response. But can't you just toggle the jumpToPDF function? I mean, if "Focus editor" is enabled, don't call jumpToPDF command. If not, why?

r-stein:
For the Sumatra part:
``` python
# ...
# removed startupinfo argument
subprocess.Popen(startCommands)

if keep_focus:
    # the command to start ST
    # could also be "C:\\Program Files\\Sublime Text 3\\sublime_text.exe"
    sublime_command = "subl"
    # time to await the start of Sumatra (in ms)
    wait_time = 500
    def focus_sublime():
        subprocess.Popen(sublime_command, shell=_ST3)
    sublime.set_timeout(focus_sublime, wait_time)
```
works surprisingly well.

PeterPablo:
Thank you @r-stein, this worked great for me. Important: Be carefull to change the leasing spaces `   ` to leading tabulator! I added an additional print statement `` to confirm on the console that this part of the code is reached.

I used **PackageResourceViewer: Open Resource** to edit `jumpToPDF.py`.

@r-stein, please consider creating a pull request. Quite certainly, this would require to make the line `sublime_command` to be smart about the executable's location though.

Thanks again!

ig0774:
[...]

PR #610, opened 2015-12-01
Keep focus for Windows and Linux

ig0774:
This is an adaptation and generalisation of @r-stein's suggestion from #495. Basically if `keep_focus` is set on Windows or Linux, it runs either `sublime_text` or `subl` to regain focus after a (configurable) short delay.

I've added some code to work out where the appropriate executable is, though it should be auto-detected on *most* configurations. The only worrying setup is ST2 where `subl` is not available on the `PATH`, in which, case the platform-specific `sublime_executable` setting must be set appropriately. Alternatively, the `sublime_executable` setting can be set and the auto-detect code will not be run.

### DIFF
--- a/jumpToPDF.py
+++ b/jumpToPDF.py
@@ -1,6 +1,7 @@
 # ST2/ST3 compat
 from __future__ import print_function 
 import sublime
+from functools import partial
 if sublime.version() < '3000':
     # we are on ST2 and Python 2.X
 	_ST3 = False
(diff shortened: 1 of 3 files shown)


jumpToPDF.py
```diff
@@ -12,10 +12,96 @@
...
+		get_sublime_executable.result = sublime.executable_path()
+		return get_sublime_executable.result
+
+	# guess-work for ST2
+	startupinfo = None
+	shell = False
+	if sublime.platform() == 'windows':
+		startupinfo = subprocess.STARTUPINFO()
+		startupinfo.dwFlags |= subprocess.STARTF_USESHOWWINDOW
+		shell = True
```
r-stein: At least in my setup this must be
``` python
shell = _ST3
```
to work with ST2 (portable)

r-stein:
I think it would be nice to add the default sublime installation path into `get_sublime_executable`, such that it works out of the box for the majority of users:
``` python
if sublime.platform() == "windows":
	if _ST3:
		DEFAULT_PATH = "C:\\Program Files\\Sublime Text 3\\sublime_text.exe"
	else:
		DEFAULT_PATH = "C:\\Program Files\\Sublime Text 2\\sublime_text.exe"
	if os.path.exists(DEFAULT_PATH):
		get_sublime_executable.result = DEFAULT_PATH
		return get_sublime_executable.result
```
*(Same for the default SumatraPDF installation path.)*

Limro:
Heads up, with those hard-coded paths. I know you said *majority*, but that is those with an english operating system.
Is it possible to ask the system for the default installation path?
I know PowerShell have two commands to find the installation paths:
```batch
$x64Path = "${env:ProgramFiles}"
$x86Path = "${env:ProgramFiles(x86)}"
```

jumpToPDF.py
ig0774: I'm happy to change it, but it works on my ST2 on Windows 7 this way. My main reason for setting `shell = True` on Windows is that it runs the command through `cmd.exe` which tends to behave more like I expect. I didn't expect it would affect the outcome here either way.

r-stein:
Thanks @Limro. Although the OS language does not seem to effect the folder. Some user might change the default directory, e.g. on an other partition.

ig0774:
Yeah... IIRC, on non-English installs there's a *hidden* `Program Files` folder and a localised hardlink to it that's localised to allow for software written by English-speakers who don't consider the possibity of a different name. Still it's (relatively) easy to makes use of the environment variables. I hadn't realised it wasn't automatically added to the `PATH`... Maybe I was thinking of ST3? In any case, I'll add in some checks.

BTW, I'm not as concerned with the default path for ST3 because the `sublime.executable_path()` function call should already tell us the executable of the version that is actually in use. It's only ST2 (where that function doesn't exist) where this becomes a problem.

ig0774:
d143719 should address the default install path on Windows. I don't think its determinable for ST2 on Linux, but let me know if I'm wrong.

A couple of comments on what I've done: 
 1. It only checks for ST2 as per my comment above. 
 1. I only used `%PROGRAMFILES%` even though `%PROGRAMFILES(x86)%` might exist. The purpose of this code is to find the executable for the current running instance of ST2. For a 64-bit version on a 64-bit OS or a 32-bit version on a 32-bit OS (eep), this will always be the equivalent of `C:\Program Files` and for a 32-bit version on a 64-bit OS it will be the equivalent of `C:\Program Files (x86)`. In other words, it will get us the right folder for the current version of ST, regardless of its bittedness.

Thanks both @r-stein and @Limro for your helpful remarks! If I've missed anything else, let me know!

r-stein:
This could also be problematic if someone uses ST2, but has ST3 in the path. E.g. if I try it on a portable ST2 it jumps to my ST3 installation (which is in the path). This will not affect many users and can easily be fixed by changing the settings, but if you have an idea to fix this.

ig0774:
Thanks. At least this should stop ST2 from running an ST3 executable.

r-stein:
Agree. If I change line `43` and line `124` to `shell = _ST3` it even works out of the box on my ST2 portable with windows.

r-stein:
Just a small remark: I think it would be more intuitive and user friendly to use seconds instead of ms as unit for the `"keep_focus_delay"` option and convert it to ms (`int(x * 1000)`).

ig0774:
Ok. I think I've addressed all of @r-stein's very good points and I added some code with some default paths on Linux, but I think I've actually found a way to make this work without a lot of fuss on ST2, since it appears that `sys.executable` points to the full path of `sublime_text` or `sublime_text.exe`, which makes sense since ST2 appears to run a statically compiled version of Python.

PR #611, opened 2015-12-01
Utility functions: "Analyze LaTeX document" and "Cache"

r-stein:
This pull request contains two functionalities. These functionalities don't effect the Enduser so far, but will (hopefully) yield new feature pull requests and improve the performance and soundness of existing features.

In short: a cache to improve performance and an analysis to retrieve a lot of information of the LaTeX document.

# Cache

As the name suggests the cache can be used to cache the results of calculations (e.g. auto-completions).
These results can either stored local (relative to tex root) or global (in the sublime cache path).
- __file position:__ There are two types of cache files:
    + _global:_ stores the file in the SublimeText cache path (ST3) or User folder (ST2)
    + _local:_ there are two possibilities for the local cache:
      * (local) creates a folder `.st_lt_cache` in the same folder as the tex_root and stores the cache files there
      * (hidden) creates a folder `$global_cache/local_cache/md5($tex_root)` and stores the cache files there
- __clear the cache:__ the cache files are deleted in the `delete_temp_files`-command together with the temporary files.
  Currently there is not other option to clear the cache, but I could implement an extra command or a periodic cleanup with a time interval.
- __cache-able objects:__ the cache uses [pickle](https://docs.python.org/3.1/library/pickle.html). Hence everything can be cached, that can be dumped by pickle (this holds for many objects)
- __usage:__ the functions `cache`, `read` and `write` can be used to access the cache.
  A cache miss is indicated by a `CacheMiss`-Exception. The intended usage is demonstrated in this code snippet:
  ``` python
  try:
      result = cache.read(tex_root, "unique_name")
  except cache.CacheMiss:
      result = calculate_result()
      cache.write(tex_root, "unique_name", result)
  # do something with result
  ```

  However the function `cache` is a short form for this:
  ``` python
  # retrieve the data from the cache and pass a pointer
  # to a parameterless function/closure to calculate it
  # if it is not available
  result = cache.cache(tex_root, "unique_name", calculate_result)
  ```

Advantages
- usually faster results. Hence enables more complex *real-time* functionalities, e.g. auto-completion.
- easy to use/small code overhead

Disadvantages
- creates a folder near the tex root (but the user can delete it with one keystroke)
- results might be outdated (but the user can manually clear the cache)

# Analysis

The analysis analyzes the `*.tex`-document and retrieves relevant information. To establish this goal the tex root and every included tex-file is analyzed. The major part of this analysis is to extract all `tex`-commands and create an object for an easy access their content. This is established by a regex-search over every document and creating an object from the result.

The result of the analysis contains:
- the tex root
- a map from each file name to its raw text content
- a map from each file name to its content without comments
- a __list of all commands__:
  each command seaches with names:
  ```
  \command[optargs]{args}[optargs2][optargs2a]{args2}[optargs3]{args3}
  ```

  for each of these names it contains the fields:
    + 'name': the string in the part
    + 'name'_region: the region of the part as a sublime region
  <br/><br/>e.g. for the `args`-part these two attributes:
    + `args`: the string in the args part
    + `args_region`: the region of the args part as a sublime region in the buffer
  <br/><br/>each command has in addition the attributes:
    + `file_name`: the name of the file, in which the entry appers
    + `text`: the full text of the entry, e.g. `\command{args}`
    + `start`: the start position in the buffer
    + `end`: the end position in the buffer
    + `region`: a sublime region containing the whole entry

## Access the analysis
The access functions are:
- `tex_root`: get the tex root
- `content`: get the content (without comments) of a document
- `raw_content`: get the content (with comments as in the file) of a document
- `rowcol`: creates a rowcol function with the same behavior as the sublime `view.rowcol` function of a document
- `commands`: returns all command entries in the file. You can apply flags for a prefiltering, e.g. remove all `\begin{}` and `\end{}` commands.
- `filter_commands`: returns all command entries in the file after applying the filter. As in `commands` it is possible to use flags for a prefiltering.

The flags for `commands` and `filter_commands`:
- `NO_BEGIN_END_COMMANDS`: exclude `\begin{}` and `\end{}` commands
- `ONLY_COMMANDS_WITH_ARGS`: allows `\com{args}` and `\com{}` but not `\com`
- `ONLY_COMMANDS_WITH_ARG_CONTENT`: only allow `\com{args}` not `\com{}` or `\com`
- `ONLY_PREAMBLE`: only in the preamble, i.e. until `\begin{document}`

The reason for the flag is create easy filters on the one side and improve performance on the other side.

# Example Usage

These are a few examples to demonstrate the information, which is retrieved in the analysis.

### Self-defined command completion
Retrieve `\newcommand`-commands in the document and insert them to the completion system similar to cwl completion.
Basic principal, (see [here](https://github.com/r-stein/LaTeXTools/blob/ana_own_command_completion/latex_own_command_completions.py) for the full file):
``` python
ana = analysis.get_analysis(tex_root)
newcommand_definitions = ana.filter_commands(["newcommand", "renewcommand"])
# the defined command is in the args part
completion_words = [c.args for c in newcommand_definitions]
```

own_command_completion

### Jump to command definition
Click on a command and open its definition in Sublime Text (I don't really think this is useful, but it demonstrates the information retrieved by the analysis)

retrieving information:
``` python
# the command is in the variable 'the_command'
# note: use 'analyze_document' to run the analysis instead of the cache
# because the regions of the cached analysis are most likely outdated
ana = analysis.analyze_document(tex_root)
newcommand_definitions = ana.filter_commands(["newcommand", "renewcommand"])
# search for the command definition entry
try:
    entry = next(c for c in newcommand_definitions if c.args == the_command)
except StopIteration:
    return  # command not found

# get the file name and the region
entry_file = entry.file_name
entry_region = entry.args_region

# open the file with ST and change the selection to entry_region
# ...
```

jumpto_command_definition

### Whole document goto overlay
Parse all section commands (with their file and region) and create an overlay, which shows those in Sublime Text.

overlay_section

### Other
this could also used to
- easily find all generated .aux files for deletion
- retrieve all labels (simplify ref/cite completion)
- IDE like rename of commands
- easy add used packages to cwl completion
- *your idea*

Advantages:
- central parsing engine -> packages with flow behavior (e.g. `subfile`, `ifthen`) can be considered in one file. The changes will propagate to every consuming function

Disadvantages:
- parsing the information might be slower, because the analysis also contains unused information (nevertheless through caching I would assume it will be faster in critical cases)

### BUG
- ~~regex regions are incorrent with windows file endings CRLF~~ (FIXED)

### TODO
- handle nested commands in the analysis (`\firstcommand{\secondcommand{}}`)
- improvements wrt control flow, e.g. check if-conditions
- ~~there should be an automatic mechanics to clear the cache (especially if the local cache is hidden)~~ (DONE via cache life span)

### Additional ideas for future pull requests
- automatically add the cache folder to the sublime ignored folders?
- parse additional information (e.g. magic header)?

I would be happy to get feedback, improvements and ideas. Especially from @ig0774.
If someone want to rewrite parts, he can post his branch in the comments and I will mostlikely merge it.

*EDIT:*

1. updated the possibility to store the local cache relative to the global cache. As @msiniscalchi suggested this option should exists
2. Better grouping of the lists, added a bug
3. fixed the bug
4. done: invalidate cache life span


KapSteR:
Amazing! This is some serious forward thinking... +1

msiniscalchi:
I love this! Definitely something I will eventually want to merge. Some
comments.

1. I am not sure I like the idea of caching stuff in the user's document
folder. For one thing, the extra stuff will be picked up by Dropbox and the
like. People are already clamoring for the option to put aux, bbl and
similar files somewhere other than the folder containing the tex and pdf
files.  But I'm open to be convinced otherwise.

2. The analyze functionality is awesome. There are two use cases I have in
mind. First, I have long wanted to refactor the get_tex_root code so as to
have one function, or at least one module, that is solely dedicated to
extracting %!TEX settings from the master file. Second, it would be great
to jump to where a given label is defined, or maybe show the neighboring
environment in a panel.

ig0774:
This is absolutely brilliant! I'm gonna have to find some time to really dig into it. I have some code I've been playing around with to try and load cwl files when new packages are found, but it got quite messy, and this might help straighten things out a bit.

A couple of things that immediately occur to me:
 1. It may be helpful to have a facility to invalidate a specific cache, i.e., if the data is known to be invalid, but it's not necessarily worth recalculating yet. For example, with the cwl thing, if a new package is added, it may not necessarily make sense to load the entries for the corresponding cwl file until the user activates autocompletions again.
 1. Just a warning: pushing code into a subfolder (as you're doing with `latextools_utils`) requires that we notify a user on upgrade that they'll need to restart ST when or if any files in that folder change. ST doesn't reload those files when their updated, only when its restarted. (A "feature" I actually abused in the plugin system for #519).

latextools_utils/cache.py
```diff
@@ -0,0 +1,218 @@
+import os
+import pickle
```
ig0774: It may be better to use cPickle here for those on ST2. It's substantially faster. Just need to do something like:

```python
try:
    import cPickle as pickle
except ImportError:
    import pickle
```

r-stein:
Thanks @msiniscalchi 

1. Arguments to keep the local cache at the TeX root folder:
  - A lot of IDEs/Editos create a local config/build/settings/workspace files. E.g. for Emacs-AucTeX for LaTeX creates an `auto` folder and eclipse TeXlipse has config folders.
  - The process is traceable by the user: he has a better insight about the creation of the cache, can manually delete it and there is not a crowing temp folder somewhere hidden
  - It does not effect the workflow on git, svn, and hg (just one more entry in the ignore folder)

  __BUT__ I agree that a lot of people use dropbox or does not want to have it for other reason. I have two suggestions, which could be set in the settings (also both could be used):
  - optionally disable the cache and always raise a CacheMiss. In this case cache consuming functions could lookup this setting and optionally disable the computation.
  - move the local cache somewhere in the sublime cache folder. We could just create an additional folder in the global cache and create a subfolder for every local cache. The name could be calculated from the usual local cache name by `md5` oder something like this.
2. At the moment magic comments `%!TEX` are not parsed and `get_tex_root` is used, hence this can not be used to refactor that part. But its designed for stuff like jumping to labels.

r-stein:
@ig0774 
to 1.:
In the current implementation you can easily invalidate a cache entry by deleting the corresponding file. It will only be recalculated when it is accessed the next time, by the person, who accessed the cache entry.
PS.: I played around with the cwl completion, cache and analysis and added the used packages in this [line](https://github.com/r-stein/LaTeXTools/blob/ana_own_command_completion/latex_cwl_completions.py#L162) and this [file](https://github.com/r-stein/LaTeXTools/blob/ana_own_command_completion/latextools_utils/used_packages.py) to retrieve the used packages.

latextools_utils/cache.py
r-stein: Thanks I will change this

latextools_utils/analysis.py
```diff
@@ -0,0 +1,426 @@
...
+        Returns:
+        A list of all commands, which are preprocessed with the flags
+        """
+        # convert the filter into a function
+        if type(how) is str:
+            def command_filter(c): return c.command == how
+        elif type(how) is list:
+            def command_filter(c): return c.command in how
+        elif callable(how):
+            def command_filter(c): return how(c.command)
```
ig0774: It might be better if this was `def command_filter(c): return how(c)` as it would enable more interesting filters to be written using all the properties of the command.

r-stein: Yes I agree and will change this.

My inital decision not to pass the whole command was based on the fear, that the caller could change the command content before creating a copy. Nevertheless I don't think that is a big issue, since the caching process is done before the analysis is returned and hence everyone else gets a "fresh" analysis. In addition the attributes are all protected and not private anyway.

latextools_utils/analysis.py
```diff
@@ -0,0 +1,426 @@
...
+        # TODO
+        # cannot pickle sublime.Region in st2 this is a hacky workaround
+        # to remove those regions if someone is interested in the regions
+        # this method is not recommended anyways
+        if not _ST3:
+            for c in result._all_commands:
+                c.region = None
+                for k in _RE_COMMAND.groupindex.keys():
+                    region_name = k + "_region"
+                    c.__dict__[region_name] = None
```
ig0774: Since `Regions` are really just glorified tuples, it should be fairly easy to provide python with the knowledge of how to pickle / unpickle them using the [`copy_reg`](https://docs.python.org/2.6/library/copy_reg.html) module. Basically we'd need this code, probably in the `except` block at the start of the file:

```python
import copy_reg

def pickle_region(region):
    return (sublime.Region, (region.a, region.b))

copy_reg.pickle(sublime.Region, pickle_region)
```

I've smoke-tested that on ST2 and seems to pickle and unpickle regions successfully.

r-stein: Thanks that is the solution ich wanted to have.

ig0774:
This might be a small issue, but in ST2, it would be good to ensure that `latextools_utils.__init__.py` imports `analysis`, `cache`, and `utils`, to ensure they are loaded with the module. As it is, by default, only `cache` is loaded---presumably because it is loaded from `delete_temp_files`, whereas `analysis` and `utils` are not available and cannot be easily loaded (so, for example, typing `from latextools_utils import analysis` or `import latextools_utils.analysis` on the console raises an exception).

Of course, this goes away once analysis is imported by any other module, but it might be as well for ensuring that its available whether or not its used in another module.

Something like this:

```python
import sublime
if sublime.version() < '3000':
    import latextools_utils.analysis
    import latextools_utils.cache
    import latextools_utils.utils
```

r-stein:
Aside: I am working on a "jump to anywhere" command, which depends on the analysis (available [here](https://github.com/r-stein/LaTeXTools/tree/jumpto_anywhere)). With this command you can click with `alt+leftmouse` and the corresponding jump will be executed. The jump depends on the context around the cursor position.
Currently supported jumps are 

- jump to references labels (`\ref`)
- jump to citation entry in bib file (`\cite`)
- open included file (`\input`)
- open included graphics with specified program (`\includegraphics`) (see https://github.com/SublimeText/LaTeXTools/pull/601)
- open documentation of a used package (`\usepackage`) (see https://github.com/SublimeText/LaTeXTools/pull/512)
- jump to self-defined command definition, i.e. jump to the `\newcommand`

r-stein:
Invalidating the cache:
To invalidate the cache I implemented a mechanism [here](https://github.com/r-stein/LaTeXTools/blob/ana_cache_invalidate/latextools_utils/cache.py).
The (simple) idea is to define the life span of the cache in the user settings and invalidate the whole cache after this life span. To do this I just store a time-stamp writing into the cache the first time. And I read this time-stamp and validate the elapsed time every time I read from the cache. If the elapsed time is greater than the life span I invalidate the cache by removing the corresponding files.

This approach is quite simple and might have minor drawbacks. However I think invalidating the cache as a whole object is a reasonable idea, because the cached object might have dependencies on other cached objects. If someone has a better idea/implementation I would be glad to see/use this. Otherwise I will use my implementation =)

ig0774:
Oooh! The "jump to anywhere" thing actually sounds really useful. And you're right that its probably best just to do whole cache invalidation: its simple and clean.

ig0774:
So I notice ([here](https://github.com/r-stein/LaTeXTools/blob/2079c6c7adc6d922046ea436bc3bc411ac1399d9/latextools_utils/analysis.py#L272)) that when the analysis is attempting to run, this code automatically triggers a save. I understand the need to get the freshest data possible, but I'm not sure if this is the way to go for a few reasons:

1. As the comment indicates, this save before open can only be done on ST3, which breaks parity with ST2.
1. This may result in flushing changes that the user doesn't want to see to the copy on disk. This may not be all that harmful if the user is using version control, but I don't want to have to make that assumption. I'm thinking about how this would be used essentially by background code while the user is editing the document, e.g., with the cwl code, typing `\` might trigger the analysis which might trigger a save. I think this causes the plugin to work in unexpected ways (and I'm convinced this, when it is merged, is going to be the backbone of a range of useful features).
1. It seems that it is unnecessary to do this in any case, since we already have the view-object in question, and can already obtain the necessary text from that.

My recommendation for all of this is to simply try to load the current data from the view. I committed a version of this to my version of the PR (see ec9f46b) which basically adds a function that loads the text for the requested file from a Sublime view if one is available and modifies `_read_file` to use that instead of reading the file from disk. I'm not really attached to my implementation---its just a proof of concept---but is there any reason we *need* to do the save or something else that I'm missing?

r-stein:
Thanks I also was not happy with the saving. I agree that your approach better and the way it should be implemented.

ig0774:
Sorry for the delayed response. I didn't really see this until now.

You're absolutely right about the order we should check things. Otherwise, I think this looks ready-to-go and will add some good functionality to the backend.

r-stein:
I made a utility function to read the file content from a view or (if not opened) from the file.

However I have problems with the `sublime` module in the utils folder, because on ST2 this overwrites the import of the normal sublime module for files in the utils folder.

ig0774:
I'm not sure I fully understand the problem. Is there a default "utils" folder in ST2? In any case, couldn't this be resolved by renaming utils to something like "file_utils"? Part of the reason that the whole folder is called "latextools_utils" was to come up with a (hopefully) unique name to deal with ST2s global namespacing of modules. I would've thought that as long as it was imported properly the "utils" name was only local to the importing module, but clearly I'm missing something...

r-stein:
I mean the file [sublime.py](https://github.com/SublimeText/LaTeXTools/blob/master/latextools_utils/sublime.py).

If I write `import sublime` in the cache this modul is imported instead of the standard sublime modul. This is only a problem for files in the `latextools_utils` folder. I think the easiest solution would be to rename this file or join it with my [utils.py](https://github.com/r-stein/LaTeXTools/blob/ana_cache/latextools_utils/utils.py) file.

ig0774:
Oh yes. That was a stupid name to choose. I'll fix that shortly.

r-stein:
I think this PR is ready to merge (except from the name of `sublime.py`).

r-stein:
It might be necessary to move getTeXRoot into the `latextools_utils` folder. 

It seems to be a problem to import a toplevel module, that imports a module from a sub-folder. I.e. importing `getTeXRoot` from `cache`, while `getTeXRoot` imports `tex_directive`. This raises a console error and files, that import the cache needs a reload to work. Hence I would suggest to move the getTeXRoot into latextools_utils and just keep a reference on toplevel for backward compatibility:

``` py
# file LaTeXTools.getTeXRoot
import sublime
if sublime.version() < '3000':
    from latextools_utils import getTeXRoot
else:
    from .latextools_utils import getTeXRoot
get_tex_root = getTeXRoot.get_tex_root
```

r-stein:
If have moved the `getTeXRoot` file in this commit https://github.com/r-stein/LaTeXTools/commit/c34fd2dc21a5d725a40cf23c749a2596a1cbc214 and would integrate it in this PR if you agree to it.

ig0774:
@r-stein: I see the interdependence given the current state of the plugin, but I think it would be better if the getTeXRoot change is a separate PR just for sanity's sake. I'm going to start working on integrating these more back-end changes after the next release.

Two less destructive options that may still get correct behaviour:

1. Remove the option to pass in a view and require the calling code to get the tex_root, although I understand the value in so far as most of the auto-completions stuff doesn't *really* need to know the tex_root, as long as it can use the analysis.

1. Something I've done to work around a similar issue (in code I haven't tried to integrate into the main branch) is just to wrap the function in a `try...except` block and reload the module to recover (see [here](https://github.com/ig0774/LaTeXTools/blob/my_master/latextools_utils/external_command.py#L36)). It's not terrible elegant, but it works.

Some other random comments:

* It should be possible to simplify `if type(tex_root) is str or not _ST3 and type(tex_root) is unicode:` this code (which appears twice in the analysis). What I've usually done is to use `basestr` on PY2. It usually looks something like this:

    ```python
if _ST3:
    strbase = str
else:
    strbase = basestring

...

if isinstance(view, strbase):
    ...
```
 It gets effectively the same result, but there are fewer things to do each time its run.

* On ST2, the default (local) cache is `$PACKAGE_DIR/User/.st_lt_cache/.st_lt_cache` which feels a bit... odd, especially as the actual data is already going to be cached in a (hopefully) unique subfolder.

r-stein:
I agree on every of your points.

r-stein:
@ig0774 Do you have plans, when to merge this? Because I have already prepared PRs, which depend on this one.

ig0774:
The immediate plan is to merge the plugin stuff and then this. Things have been moving slower here than I intended because of the interference of real life.

r-stein:
okay thanks =)

ig0774:
@r-stein: Just to let you know, I've included this is the most recent release

excetara2:
@r-stein Just curious if you have implemented `Whole document goto overlay` in a different pull request. I was wondering about this for books and thesis as I recently noticed that goto Symbol in Project doesn't do anything for LaTeX projects.

excetara2:
@r-stein Also a good use of this I think would be to enable partial building because I don't think that is currently built into `LaTeXTools`?

Just by inserting an `\includeonly{...current_file...}` just before the `\begin{document}` of the root file to enable faster building of large projects.

r-stein:
@excetara2 

1. the `Whole document goto overlay` is still only offline on my pc, but I am looking forward to create a PR for it.
2. I am struggeling with the advantages of supporting `\includeonly` in the analysis (except from the performance). Since the user is most likely interested in the whole document. On the performance side it will be way more effective to somehow cache the result per tex file and only analyze each file if it has been changed.

excetara2:
@r-stein 
1. Okay awesome I look forward to it.
2. My thinking for a partial build is somewhat different and should just be an option to toggle (like focus) when the user knows they are only working on one section that would then  `\includeonly` the current view. The latex build system then only compiles the tex file for the sections in `\includeonly` and uses the `.aux` files already generated for all other sections so references/cites still work properly.  
Maybe I just realized this is what you mean about possibly cache the result per tex file? Essentially using the analysis it always performs a check if the files are different before recompiling the `.tex` files for all included files in the project. This might be an option but still often if your working on a large document typically the view of the current file your working on or looking at is the only one you want compiled at the current moment. As stated, I think it should be a toggle and not the default option but it can drastically increase compile time for some large documents. I've mainly had problems with large documents (thesis/book) that have many image files that take a long time to compile so I've used partial building within the current file your working in the past in `emacs`.  
Anyways, it should be easy to implement so I might have a go at this when I get a chance just. I think you should just be able to insert the command `includeonly{currrent view}` at the end of the `PREAMBLE`.

PR #616, opened 2015-12-04
Pluggable builders

ig0774:
This pull request ports the builder system to using the plugin system shared with #519. For a description of the plugin system, see #618.

Right now, this doesn't add any functionality, except that for supporting storing custom builders in the `User` package with the extension `.latextools_builder`. At the same time, it does not take any away.

The intention of these changes is as a building-block to enable builders to take advantage of library functions in the LaTeXTools code base. 

I have two use cases in mind for which this is a stepping-stone:
 1. Extract the engine-selection code and `%!TEX option` parsing code from the TraditionalBuilder to make it available to other builders and other parts of the LaTeXTools code, if desirable.
 1. Using the plugin system makes it possible for other parts of the LaTeXTools code to use the builders in non-building contexts. In particular this would enable us to do *something* with the (currently unused) `cleantemps()` function that could be leveraged by the `delete_temp_files` command.

This could, of course, prove useful for other purposes as well.

With these changes, builders can be accessed like so:

```python
builder_class = get_plugin('traditional_builder')
builder = builder_class(tex_root, output, builder_settings, platform_settings)
...
```


msiniscalchi:
@ig0774 this, as well as #519 and #611 , requires a new branch. The changes are way too extensive. So here's my plan.

1. RIght now my priority is to ttransition LaTeXTools to a tags-based release system for Package Control to pick up. Right now I'm using branch--based releases, which are bad in many ways and now deprecated. I hope this can happen without disruption to the users, but certainly I need to keep things simple in the code base and not add any new features or deep plumbing.

2. I love the idea of a generic plugin system. Indeed my idea for multiple pdf viewer support was to implement something like this. I need to study this better, but it does seem to be the way to go.

3. Ditto for the helper functions.

Thank you thank you thank you @ig0774 and @r-stein !

msiniscalchi:
@ig0774 a further comment. I understand this is potentially painful, but ideally it would be nice to have three separate pull requests:
- one for the basic plugin system
- one for the builder plugins
- one for the cite completions.

This way I would create a branch with the basic plugin system, and code that uses it would have to go to that branch but wouldn't have to include the plugin system itself. Once we can be reasonably sure that there is no breakage, we can merge the branch to master.

ig0774:
@msiniscalchi: That's actually no problem. I already have those three separated into different branches in my repo. #618 is just the bare plugin system. Hopefully this will make it clearer what the changes are.

PR #617, opened 2015-12-04
Script Builder

ig0774:
This pull request adds an actual functioning script builder, at least as far as I understand the intention of it. It is able to run simple lists of commands, configurable in the `LaTeXTools.sublime-settings` file or even launching external build tools. There's some fairly extensive docs added to the README on how to configure it, etc., and some warnings on the assumptions that LaTeXTools makes.

The builder works by constructing and yielding `Popen()` objects to the `make_pdf` command, which has been modified to accept `Popen` objects *in addition to* strings or lists.


ig0774:
I realise this needs to be updated. I also want to change the setting we get the list of commands from from `command` to something like `script_commands` to remove any ambiguity between the commands for the script builder and the `command` setting that drives the traditional builder.

PR #618, opened 2015-12-05
Plugin System

ig0774:
Here's the plugin system shared by #616 and #519. It consists of two files:

* `latextools_plugin`: the actual plugin system
* `latextools_plugin_internal`: this is used as a point to store the global registry and other variables that should be set once.

**Plugin system:**
Essentially the module `latextools_plugin` contains some code for plugin auto-discovery using a user-configurable path and a plugin marker class (`latextools_plugin.LaTeXToolsPlugin`). It provides a named-based registry of plugins that can be accessed by LaTeXTools code and is intended to make very few assumptions about the plugin code so that the registry can be used to provide extensions at arbitrary points in the existing LaTeXTools code.

I realise that's kind of abstract, so here's an example:

*Plugin:*
```python
from latextools_plugin import LaTeXToolsPlugin

class MyCustomPlugin(LaTeXToolsPlugin):
    # implementation goes here
    # the functions to be implemented depend on what code will use this plugin
    def do_something(self):
        pass
```

*Plugin Consumer:*
```python
import sublime

if sublime.version() < '3000':
    from latextools_plugin import get_plugin
else:
    from .latextools_plugin import get_plugin

plugin = get_plugin('my_custom')
# instantiate and run the plugin
plugin().do_something()
```

Further documentation on the plugin system is in the doc-string at the top of `latextools_plugin.py`.

*Technical Note*: The `latextools_plugins_internal` part is needed to work around some issue with ST's loading mechanism. In particular, without storing this data here, it tends to get erased on reload and we lose track of the requested plugin paths, etc.


msiniscalchi:
This is perfect, thanks! I have created a branch called "plugin" and will merge your pulls there, so I can play with them.

Since you have contributed so much to LaTeXTools, and you seem to have the time and energy, would you like to join as a co-developer? In practice this means that you could develop using the official repo.... and, I hope, also help deal with user requests / bug reports. Happy to take this offline to discuss---it's a bigger commitment than you may think...

ig0774:
@msiniscalchi Yes, let's discuss offline

msiniscalchi:
@ig0774 great; I don't see your email address in your profile. You can write to my gmail address (same as as my github handle).

msiniscalchi:
OK, I pulled this into the "plugin" branch. I'll start studying it now!

ig0774:
Closing this as all further developments will happen in the main `plugin` branch.

PR #621, opened 2015-12-11
CWL completion for environment names

r-stein:
Changes the cwl completion inside the `\begin` and `\end` command to provide the environment names.
cwl_env_completion

Integrates into fill-all helper:
cwl_env_fillall_completion

Technical detail:
The environment names are extracted from the `\end{env}` lines in the cwl file, because this directly generates a list without duplicates. This should be adequate, because every environment needs the `\end`-command.

*Edit:*
Added Fill-All Helper

### DIFF
--- a/Default (Linux).sublime-keymap
+++ b/Default (Linux).sublime-keymap
@@ -68,6 +68,15 @@ LaTeX Package keymap for Linux
 				"operand": "(?:\\\\include)|(?:\\\\input)|(?:\\\\includegraphics(?:\\[[^{}\\[\\]]\\])?)|(?:\\\\addbibresource(?:\\[[^{}\\[\\]]\\])?)|(?:\\\\bibliography)|(?:\\\\documentclass(?:\\[[^{}\\[\\]]\\])?)|(?:\\\\usepackage(?:\\[[^{}\\[\\]]\\])?)|(?:\\\\bibliographystyle)", "match_all": true}],
 		"command": "latex_fill_input", "args": {"insert_char": "{"}},
 
+	{	"keys": ["{"],
+		"context":  [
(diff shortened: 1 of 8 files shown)


ig0774:
I love this feature, as with all your pull requests, but (as always) I've got a few niggles:

First, as is the environment completion makes a hard assumption that the `\begin{}` or `\end{}` command is the only thing prior to the completion on the line, so completions work in the screenshot, but not in a situation like this:

```latex
some text: \begin{}

\end{}
```

Which is why, generally speaking, the completions are implemented using backwards regexes.

Second, without the quick panel, this feels a bit out of place, given the rest of the completion features,  which isn't to say that this is wrong, just that we should probably have a quick-panel feature as well.

Third, (building on the above) the environment completions should be integrated into the FillAll feature.

r-stein:
Thanks (1) is fixed.

To (2) and (3):
I agree there should be a quick panel for this and it should hence be integrated into the fillall feature (and I will do this). However it might be better to disable the quick panel by default, because of the missing completeness of the environment names. Since the used packages are not parsed (yet) and not all cwl files are available.

ig0774:
On (2) you're right. Even if we *did* parse all the cwl files (I'm actually making that happen for other reasons), and compare that against the loaded packages, we'd likely be missing some environments because the cwl files themselves are incomplete. (TeXStudio, for example, has a much more complete set of cwl files than the ones in the LaTeX-cwl package and even they don't cover every package or option). Still, I think having *something* appear if the FillAll command is invoked is probably a worthwile goal.

BTW, you don't have to feel obligated to implement that. I'm happy to do it.

r-stein:
Now it is available, but is not triggered by default.

And parsing the used packages/cwl files will be easy with the analysis https://github.com/SublimeText/LaTeXTools/pull/611.

ig0774:
Thanks!

PR #622, opened 2015-12-12
Useful Default Settings

ig0774:
Per @r-stein and @msiniscalchi's conversation on #601, this renames `LaTeXTools.default-settings` to `LaTeXTools.sublime-settings` so that newly-added settings will have default values (this mitigates a certain class of bugs). It also provides a menu option to open the default settings file, but in read-only mode.

A couple of things:

1. In my view, this does not obviate the need to run the migration. Almost certainly the build settings need to be tweaked, and the easiest way to do this is migrate the default settings. We still need to update the migration so that if there already is a user-specific `LaTeXTools.sublime-settings` file, we don't nuke their custom-configured options.

2. There are probably more drastic option we could do to make the package sublime-settings file unwritable (mark it read-only at the filesystem level, create a listener in ST to always set it read-only). I've taken the minimalist route: when you open it through the menu, it is read-only. I think that should suffice?

3. I've done this as a pull request a) to elicit comments and b) because merging this will break any existing pull requests that change the default settings file.


msiniscalchi:
Thanks @ig0774 . I'm still thinking about the migration issue. The approach @r-stein proposes, which is what ST does by default and this patch implements, has the advantage of being very flexible. For expert users, this is ideal. However, many users really don't want to have to deal with complexities---they want sane defaults. For these users, the only options they really need to worry about are the paths to the viewer and tex binaries. However, both TeXLive and MikTeX put themselves on the path these days. Sumatra doesn't, but we tell users to either do so, or set the relevant option. Even that is something we could easily avoid, by looking in both `C:\Program Files\SumatraPDF` and `C:\Program Files (x86)\SumatraPDF` by default.

On the other hand, I think the real value of migrating the entire settings file (while preserving the user's defaults) is in the *comments*. But here, too, the issue is subtle. My original thought about migrating the whole file was that, since comments were copied to the user's settings as well, there was no need to consult the default settings file at all. Hence, that file was hidden from the user (by not exposing it via menus, and by changing its extension). What complicates matters is that the *user* may also add comments to his/her own version of the settings file---to remind himself/herself of why certain options are set in a certain way. This makes writing a good migration solution harder. 

This is why I am drawn more and more to the ST way. We simply don't touch the user's file at all (except of course if the user asks us to generate a fresh settings file from the default one). 

That said, @ig0774 is absolutely right; when we pull this, we need to make sure we don't break anything recent (or,more precisely, we need to resolve all conflicts as they arise).

r-stein:
I agree on your points and I don't think adapting existing pull requests will be a big issue.

Since you implemented a command to make the default settings read only, it might be worth a thought to also implement the user settings command. This commands should open the user file if it exists and else migrate and open it.

ig0774:
Its pretty straightforward to implement that, but I wonder if it's the right thing to do. In my mind, one of the nice things about the ST way of handling settings is I can a) have sane defaults and b) have a clear record of what I have changed for my particular environment and c) (assuming the defaults are sane) only have to worry about options or features if I become interested in them. 

Hence, I'm wondering if tying the view settings option to the migrate command doesn't *negate* some of the benefits of moving towards something closer to the ST model for settings. I'm not saying we get rid of the migrate command, just maybe not force it on people.

msiniscalchi:
@r-stein I tend to agree with @ig0774 on this. BTW there already is a "Settings - User" menu item, so only a "Settings - Default (RO)" command is necessary, and the migration command can be retained, but as a separate item.

r-stein:
I just think, that a "Your personal settings does not exists yet. Do you want to migrate the LaTeXTools settings first?"-message is better than showing an empty file. As I said only migrate it if the file does not exists and I also think we should ask the user first.
The migration command also migrates the settings if they exists, hence it should stay in this scenario.
However if you don't like that idea, I am perfectly fine with that =).

msiniscalchi:
@r-stein I see, sorry for misunderstanding. I thought the proposal was to remove the migrate command altogether.

ig0774:
@r-stein: Actually, I like the idea of a prompt, so I've implemented that in 5aa8fa1. I guess I thought you meant just running the migration.

Now, when selecting `Settings - User`, if no settings file exists it will prompt you to migrate the settings. If you select "Ok" (we could change this) it runs the migration. If you select "Cancel", it gives you a blank settings document.

Sound good to everyone?

Also, right now the prompt says:

> You do not currently have a LaTeXTools.sublime-settings file.

> Do you want to run the migrate command to create a copy of the default settings file in your User directory?

But I'm not entirely happy with that text. Any better ideas?

msiniscalchi:
@ig0774 great! As for the text, how about:

  You do not currently have a personalized LaTeXTools.sublime-settings file.

  Create a copy of the default settings file in your User directory?

[Then again, I'm the guy who came up with "Texification succeeded," so you may not put too much stock in my prose...]

ig0774:
@msiniscalchi: I like it better than mine!

msiniscalchi:
@ig0774 are you OK with me pulling this into master, so I can play with it for a couple of days before we release?

ig0774:
Yes, I'm happy it works...

PR #623, opened 2015-12-12
Use packages to determine which cwl files to use

ig0774:
This is a pretty drastic overhaul of the cwl code to support loading completions based on the packages used in the current document. I think this is a nice-to-have feature, since it requires the user to mess around less with the settings. This behaviour can be *disabled* by setting the `cwl_autoload` setting to `False`.

The packages are parsed from the main file, using the analysis tools. and *added* to the list in the existing `cwl_list` setting. This is because certain default files (`tex.cwl`, `latex-209.cwl`, `latex-document.cwl`) which should probably be loaded are otherwise missed.

The other key difference is that we now only parse the cwl files once and store the results of that for future use. This is done in a background thread when the plugin was loaded. This is done to keep the completions relatively fast (the naive approach where all the parsing is done on a single thread is fine for a relatively short list of completions, but bogs down quickly when there are a large number of cwl files loaded).

The downside to this approach is that if new cwl files are added or current files modified, those changes won't be parsed until the plugin is reloaded.

A side-effect of these changes is that it is now possible, on ST3, to create an exploded `LaTeX-cwl` package directory, which will override the `.sublime-package` version. Previously, only the `.sublime-package` version would be used. This could be useful for advanced users who want to add their own completions.

### DIFF
--- a/latex_cwl_completions.py
+++ b/latex_cwl_completions.py
@@ -456,7 +456,12 @@ def cwl_parsing_handler(callback):
 # whether it is in a .sublime-package file or an expanded directory
 if _ST3:
     def get_cwl_package_files():
-        results = sublime.find_resources('*.cwl')
+        results = [
+            r for r in
+            sublime.find_resources('*.cwl')
+            if (r.startswith('Packages/User/cwl/') or
+                r.startswith('Packages.LaTeX-cwl/'))
+        ]
         return(results, True) if results else ([], False)
 else:
     def get_cwl_package_files():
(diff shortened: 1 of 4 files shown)


latex_cwl_completions.py
```diff
@@ -110,118 +285,251 @@ def on_query_completions(self, view, prefix, locations):
...
+# sublime-package file or an exploded directory; returns a tuple
+# consisting of the list of cwl files and a boolean indicating
+# whether it is in a .sublime-package file or an expanded directory
+if _ST3:
+    def get_cwl_package_files():
+        results = [
+            r for r in
+            sublime.find_resources('*.cwl')
+            if (r.startswith('Packages/User/cwl/') or
+                r.startswith('Packages.LaTeX-cwl/'))
```
r-stein: This should be `Packages/LaTeX-cwl/`

ig0774: Good catch!

Issue #624, opened 2015-12-13
Can't open image with shortcut C-l-o in v3.6.0


Gabriel-p:
I'm running elementary OS 0.3 (Freya) and I use [Viewnior](http://siyanpanayotov.com/project/viewnior/) as my default image viewer.

I added the following to my settings file:

````
// ------------------------------------------------------------------
// Opening files included into the tex source code
// ------------------------------------------------------------------

	// image types you use in latex
	// these types will be used for autocompletion and 
	// opening of included images, when no extension is written
	"image_types": ["png", "pdf", "jpg", "jpeg", "eps"],

	// the commands to open image files. The extensions will be matched from top to bottom.
	// If no extension is given, all extensions will be matched.
	// If there is no match the image will be opened with Sublime Text.
	// the extension:
	//		can either be a string for one extension
	//		or an array of string to match more than one extensions.
	// the command:
	//		can be the name to start the programm. The image file will be passed
	//			as the argument.
	//		can be a string as a command with args:
	//			e.g. "open -a PixelMator", in this case the file will be appended
	//			if the position is relevant use "$file" in the string
	//			e.g. "open -a PixelMator $file"
	//			this will be substituted by the path to the file
	"open_image_command": {
		"linux": [
			{
				// open pdf documents with evince
				"extension": "pdf",
				"command": "evince"
			}
			// uncomment these lines to open all other images with the default programm
			,
			{
				"command": "viewnior"
			}
		]
	},
````

but this just opens the app with no image. I tried `"command": "viewnior $file"` but it does the same thing.

ig0774:
Can you grab the console output from when you press `C-l-o`? Also, which version of ST are you using?

Gabriel-p:
This is the console at first load (I'm using Sublime 3, build 3083):

````
DPI scale: 1
startup, version: 3083 linux x64 channel: stable
executable: /opt/sublime_text/sublime_text
working dir: /home/gabriel
packages path: /home/gabriel/.config/sublime-text-3/Packages
state path: /home/gabriel/.config/sublime-text-3/Local
zip path: /opt/sublime_text/Packages
zip path: /home/gabriel/.config/sublime-text-3/Installed Packages
ignored_packages: ["Vintage"]
pre session restore time: 0.321052
loading dictionary Packages/Dictionaries/English (British).dic
startup time: 0.734682
first paint time: 0.745848
reloading plugin Default.block
reloading plugin Default.comment
reloading plugin Default.copy_path
[...]

PR #626, opened 2015-12-14
Bugfix: Remove subprocess.Popen shell-argument in the "jump to image file" command

r-stein:
This fixes https://github.com/SublimeText/LaTeXTools/issues/624. It just removes the shell argument in the `subprocess.Popen`-command for opening image files, because it seems not to be necessary and produces errors on linux.
*(I added this argument before splitting the argument string into an argument array.)*


msiniscalchi:
Indeed this is needed on OS X as well (at least with El Camino and probably other recent versions). Thanks for the quick fix!

PR #628, opened 2015-12-15
Delete duplicate lines in default-settings.

Josef-Friedrich:
The settings  "tex_file_exts" and "latextools_set_syntax" are listed twice in the main settings file (LaTeXTools.default-settings).


ig0774:
Thanks for catching that... Result of too much merging.

PR #633, opened 2015-12-27
Enable overriding settings with platform-specific settings

ig0774:
This is similar in spirit to #609 and completely compatible with those changes, but introduces some new changes into how settings are loaded. For the motivation for doing this, see #412.

Basically, the idea is to generalise the OS-specific settings stanza and the OS-specific `builder_settings` stanza to all or most of the other settings, with the intention that the user can specify *platform-specific* versions of any setting and have those honoured.

From an API perspective, this means that its usually only necessary to call `get_setting()` and it will load the appropriate platform-overridden setting if it exists or otherwise continue to work as it does in the current release.

Among other things, this is meant to clear up the ambiguity discussed in #412, so that it's possible to specify either the `command` or `program` setting for the traditional builder in an OS-specific manner while still honouring any existing settings that are configured.

To do so, it constructs a hierarchy of settings. I'll use an abbreviated versions of settings files to demonstrate with numbers indicating the priority in which a setting is use (1 being the highest priority and 6 being the lowest and the platform assumed to be osx):

**`LaTeXTools.sublime-settings`**
```javascript
{
    ...
    "osx":
    {
        "builder_settings":
        {
            "program": "pdflatex" // #5
        }
    },
    ...
    "builder_settings":
    {
        ...
        "program": "pdflatex" // #6
        "osx":
        {
            "program": "pdflatex" // #4
        }
    }
}
```

**Project File**
```javascript
{
    ...
    "settings":
    {
        "osx":
        {
            "builder_settings":
            {
                "program": "pdflatex" // #2
            }
        },

        "builder_settings":
        {
            "program": "pdflatex" // #3
            "osx":
            {
                "program": "pdflatex" // #1
            }
        }
    }
}
```

These changes work with every current feature in LaTeXTools, including the `open_image_command` settings (which uses platform-specific settings, albeit in a different way), but there is some potential that this change could cause settings to behave in unexpected ways for future code, so I've left the current implementation of `get_setting()` in under the name `get_raw_setting()`.

Right now, I'm not sure there are a whole lot of use-cases for this feature, but with the script builder and some similar changes to allow viewer plugins to be created that I'm working on, something like this is becoming more necessary, I think.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -479,6 +479,26 @@ This sets `main.tex` as the master tex file (assuming a multi-file project), and
 **Note:** tweaking settings on a project-specific level can lead to even more subtle issues. If you notice a bug, in addition to resetting your `LaTeXTools.sublime-settings` file, you should remove at LaTeXTools settings from your project file.
 
 
+### Platform-Specific Overrides ###
+
+All of the generally available settings can be overridden for a specific platform in either a project file or the main `LaTeXTools.sublime-settings` file. This allows you to, for example, keep many of the same settings across platforms, but customize any settings on a platform-specific basis. This is particularly useful with builder settings. Previously, some settings could be overridden on a platform-specific basis, but this required extra code to provide this option.
+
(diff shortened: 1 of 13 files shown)


ig0774:
Per out of band conversation, these changes are too extensive and probably not the best idea, so I'm withdrawing them.

PR #636, opened 2016-01-04
Common `%!TEX` directive parsing

ig0774:
Some back-end changes that do two things:

1) It moves all the parsing code for `%!TEX` directives to a single place

2) It extracts the code for determining the `engine` and `options` settings used by the traditional builder and makes them available to all builders

### DIFF
--- a/latextools_utils/tex_directives.py
+++ b/latextools_utils/tex_directives.py
@@ -46,65 +46,71 @@ def parse_tex_directives(view_or_path, multi_values=[], key_maps={},
     '''
     result = {}
 
+    is_file = False
     if isinstance(view_or_path, sublime.View):
         lines = view_or_path.substr(
             sublime.Region(0, view_or_path.size())).split('\n')
     elif isinstance(view_or_path, strbase):
         try:
-            with codecs.open(view_or_path, "r", "utf-8", "ignore") as f:
-                lines = f.readlines()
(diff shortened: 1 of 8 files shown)


latextools_utils/tex_directives.py
```diff
@@ -0,0 +1,116 @@
...
+import sys
+import traceback
+
+if sys.version_info < (3, 0):
+    strbase = basestring
+else:
+    strbase = str
+
+TEX_DIRECTIVE = re.compile(
+    r'%+\s*!(?:T|t)(?:E|e)(?:X|x)\s+([\w-]+)\s*=\s*' +
```
r-stein: 1. whitespaces between `%!` and `TEX` should be allowed, hence `%+\s*!\s*(?:T|t)` instead of `%+\s*!(?:T|t)`
2. You could use the `re.IGNORECASE` flag instead of writing every letter und upper and lower case

ig0774: Re 1. It's implemented this way for consistency with [Texworks](https://github.com/TeXworks/texworks/blob/master/src/TeXDocument.cpp#L3033) and TeXshop, which understand `%` as a comment and `!TEX` as the actual directive.

Re 2: Yes. This is an artefact of trying to use the same re both here and [in the Atom version](https://github.com/msiniscalchi/atom-latextools/blob/master/lib/parsers/tex-directive-parser.coffee#L3). I suppose I could've used flags in both places.

PR #638, opened 2016-01-05
Change default value for `hide_build_panel` to `no_warnings`

r-stein:
This PR just changes the default settings for `hide_build_panel` from `never` to `no_warnings`, because this seems to be the better default value.
I uses the default value `never` to not change the current behavior unless explicitly wished. However on the hindsight most users seems to wish to hide the build panel and does not read all settings.
The value `no_warnings` seems to be reasonable, because without warnings the build panel only contains the file name, build engine and that the build is succeded. The success can also be indicated by hiding the build panel and the other information is redundant in every build.

### DIFF
--- a/LaTeXTools.sublime-settings
+++ b/LaTeXTools.sublime-settings
@@ -49,7 +49,7 @@
 	// "no_errors" (only hide the panel if the build was successful even with warnings),
 	// "no_warnings" (only hide the panel if no warnings occur) and
 	// "never" (default, never hide the build panel)
-	"hide_build_panel": "never",
+	"hide_build_panel": "no_warnings",
 
 	// valid texfile extensions
 	"tex_file_exts": [".tex"],


ig0774:
I'm going back and forth on whether or not this is a good idea. On the one hand, I see the appeal of this as a default, and our settings file is rather intimidating. Moreover, it seems to work well for most purposes and if there are no warnings, there isn't a lot of need to display this (especially as the "Build succeeded" text is actually a bit misleading; it only indicates that no Python errors occurred—we get quite a few reports of where nothing appeared to happen on a build because of an error on the command line that our builders don't pick up on since we don't do anything with exit codes). On the other hand, there is an edge-case when the builder exits quickly, e.g. when latexmk decides it has no work to do, at least for me, the console never appears at all (or appears and vanishes before I can see it). I worry that it may appear to someone that nothing happened at all in the case where the build failed due to some misconfiguration (usually a PATH setting), but we don't pick up on it because nothing was written to the log.

In short, I think this is a good idea in principle but we need some better error handling before it's a good idea in practice.

ig0774:
I'm closing this since I think its covered by #732. Hope that's ok.

PR #639, opened 2016-01-05
Pluggable viewers

ig0774:
This leverages the plugin system from #618 to allow for user-defined viewers. In addition to moving the viewing code to plugins, it add several new features.

**New Features:**

 * A fully customisable viewer called `command` that allows any running arbitrary commands to open PDF documents. It should work for any application that exposes a sane command-line interface. As such, it supercedes #532. It probably doesn't support Zathura out of the box, since that appears to need some extra handling, but for now, something like [the shellscript here](https://bbs.archlinux.org/viewtopic.php?id=189071) could be modified to add support.

 * A viewer for `Okular`. This is actually quite easy to support.

 * A viewer for `Preview.app`. This isn't used by default, but is available. It is included a) as a very simple demo of what a viewer-plugin might look like and b) because the `focus_st()` hack doesn't really work on OS X as it does on other platforms.

No functionality is changed without modifying settings, however.


czechboy0:
I'm impatiently waiting for this, what's the approximate timeline of this being merged?

ig0774:
@czechboy0: it should be in the next major release. I don't have a more specific time-frame than that.

emanuelealiverti:
tested with qpdfview 0.4.16.99 on linux, perfectly working.

The following configuration is needed:
**Forward Seach**

    "viewer": "command",
		"linux" : {
			"forward_sync_command":"qpdfview --unique $pdf_file#src:$src_file:$line:$col",
			"view_command": "qpdfview --unique"
		}
(syntax is a bit different from okular)

**Inverse Seach**
you have to set this variable in `~/.config/qpdfview/qpdfview.conf`, `[documentView]` section:
`sourceEditor=subl %1:%2`

(hope it will be useful for someone else, posted here because I don't know where it would be more appropriate)

r-stein:
You could create a new [wiki page](https://github.com/SublimeText/LaTeXTools/wiki) for viewers. (everyone can edit the wiki)

emanuelealiverti:
Great, done it! Should I delete the comment now?

r-stein:
If you wish to, but that is not necessary.

r-stein:
shouldn't the view command be `"view_command": "qpdfview --unique $pdf_file"`?

emanuelealiverti:
I'v just adapted the [Okular eaxmple](https://github.com/SublimeText/LaTeXTools#command-viewer). `C-l,v` command works properly

r-stein:
The view command should also open the pdf in the viewer, if the viewer is closed. We may also change the readme.

ig0774:
@r-stein: The command viewer automatically appends `$pdffile` to the command if no variables are used in the command. The idea being that it's pretty useless to call a PDF viewer without the name of the file at least.

r-stein:
@ig0774 Ah that makes sense. I should have reread that readme section =)

Issue #640, opened 2016-01-07
Close completion panel if "\\" is inserted


FichteFoll:
Coming from LaTeXing (which was unusable due to https://github.com/LaTeXing/LaTeXing/issues/198), I was really used to the auto-complete panel automatically getting closed once I inserted the second backslash and I missed it immediately.

r-stein:
Done in https://github.com/SublimeText/LaTeXTools/pull/641

FichteFoll:
:+1:

PR #641, opened 2016-01-08
Hide (cwl) auto-completion popup if the leading backslash is escaped

r-stein:
This resolves https://github.com/SublimeText/LaTeXTools/issues/640.

Since `\` is in the `auto_complete_triggers` the auto-completion also appeared if it was escaped. This PR checks whether it is escaped and optionally hides the auto-completion. If the user has not configured the auto-completion to popup in the scope, then the auto-completion is not hidden.

It also removed the cwl entries from the auto-completion if the leading backlash is escaped.

lt_escape_cwl

### DIFF
--- a/latex_cwl_completions.py
+++ b/latex_cwl_completions.py
@@ -19,14 +19,15 @@ else:
     from .getRegion import get_Region
     from .latextools_utils import get_setting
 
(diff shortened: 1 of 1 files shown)


ig0774:
Thanks @r-stein!

PR #649, opened 2016-01-30
Adding includesvg to input completion

Eloo:
Simply added the \includesvg regex and svg_filter to latex_input_completions.py


ig0774:
Merged. Thanks for the contribution! How would you like to be credited?

Eloo:
Thanks, but there is no need to be credited for this tiny PR :)

Issue #647, opened 2016-01-27
Set spellchecking language in the document


fph:
Feature request: it would be cool if Latextools honored Texstudio's `% !TEX spellcheck = it-IT` directives to set the spellchecking language.

r-stein:
I think this is a nice idea and I am preparing a PR. Just a few points:
- Do you know if the `spellcheck` directive only holds for the current document or should also extracted from the root document?
- I am not sure on which events the function should be triggered. E.g. on every save would directly react on changes, but slow down ST.
- It seems only to be useful if the [Dictionaries](https://packagecontrol.io/packages/Dictionaries) package is installed. (In addition translation locales to the dictionary language names cannot be done in a generic way and requires a full mapping.)

fph:
Thanks for the enthusiastic support!

1. No, I don't know. It would make sense to me if it held only for the current document, though. Also because I can imagine use cases where different `\include`d files use different languages.

2. Good point. AFAIK it is common for this kind of directives to work only if they are in the first 10 lines of the file (the only reference online I could find, though, is [here](http://tex.stackexchange.com/a/46218/1371). Would it be acceptable to scan on every save only the first 10 lines to see if they start with `% !TEX`? (Incidentally, I see that LaTeXTools currently handles the `%!TEX root=...` directive in `getTeXRoot.py` by scanning only up to the first non-comment line.)

3. I agree with you that it seems necessary to have Dictionaries installed, and that a manual locale-to-language-name mapping seems necessary.

4. Incidentally, [this tex.stackexchange answer](http://tex.stackexchange.com/a/84687/1371) claims that Sublime Text already supports spellcheck directives, but I couldn't find it confirmed anywhere else. I have tried it with `it_IT` and it doesn't work, so I assume it is a mistake in the answer.

r-stein:
- I think it would be nice to first check the current file and if it is not defined in the root
- My current implementations checks whether the dictionary packages is installed and creates a mapping from the locales to the names. In addition the user can define his own mappings in his settings (He might have own dictionaries or other dict packages). If someone is interested, the implementation is available [here](https://github.com/r-stein/LaTeXTools/blob/autodetect_spell_check/detect_spellcheck.py).
- The answers states it understands `root` and `program` but not `spellcheck`, which is truee

ig0774:
@r-stein: Is there anything apart from #636 that prevents this from being a PR? This looks like a nice feature to add.

r-stein:
[...]

PR #652, opened 2016-02-07
Autodetect spell check

r-stein:
This PR depents on https://github.com/SublimeText/LaTeXTools/pull/636 (hence that should me merged first). It has been discussed in https://github.com/SublimeText/LaTeXTools/issues/647.

This PR parses the `%!TEX spellcheck` directive to automatically change the dictionary for the spell-checker integrated in ST. It supports the [Dictionaries](https://github.com/titoBouzout/Dictionaries) package and the user can define his own mappings, if he has additional dictionary packages / dictionaries. We need explicit mapping, because the translations are not straight forward.


ig0774:
Merged. Thanks so much for this!

Issue #655, opened 2016-02-17
COULD NOT COMPILE! error when trying to comile on Linux


kuzmamar:
I am not able to make this work. When I try to run compiling using Ctrl b from sublime I get:

> TraditionalBuilder: Invoking latexmk... 
> 
> COULD NOT COMPILE!
> 
> Attempted command:latexmk -cd -e -f -pdf -interaction=nonstopmode -synctex=1 min.tex
> Build engine: Traditional Builder

But running the command from terminal works fine.
The problem may be in file makePDF.py between 101 - 174. Some kind of exception is thrown.

ig0774:
Is there any output in the Sublime console when this happens? You can get to the Sublime console either by pressing **ctrl + `** or the menu option **View | Console**.

 Normally this sort of thing is caused by a configuration issue e.g. where the path of your terminal doesn't match the path available to ST. Also what platform are you on? And which version of ST?

kuzmamar:
Thanks @ig0774. The console output helped. But I am still confused. I have installed texlive on Linux Ubuntu 14.04 and in my .bashrc file i apended the installed folder "/usr/local/texlive/2015/bin/x86_64-linux" to the PATH variable. For some reason latextools could not find latexmk command. But then I tried to append the "/usr/local/texlive/2015/bin/x86_64-linux" to PATH variable in "texpath" under "linux" in user settings. This worked, which made me think under what user is the command executed? Certainly not me, because it didn't have my PATH variable available.
I am using sublime text 3.

r-stein:
If you don't start ST from the shell, it won't read the environment variables. (The compile command is not run in a shell.)

I think variables set in the `~/.profile` are also access-able in ST.

ig0774:
It's as @r-stein says: The problem is the environment visible to graphical applications differs somewhat from the environment visible on the login shell. Basically the environment graphical applications get derives from the display manager (on Ubuntu, usually `gdm`) which normally uses `/bin/sh` rather that `/bin/bash` to initialise its environment.

[The Ubuntu Wiki](https://help.ubuntu.com/community/EnvironmentVariables#Persistent_environment_variables) has a good run-down of the various places that you can put environment variables so they are visible to graphical programs, but basically , `~/.profile` is  the lowest common denominator read by both X and (almost all of) the various shell programs.

kuzmamar:
Thanks @r-stein @ig0774 for clarifying that.

YerongLi:
[...]

PR #656, opened 2016-02-19
System Check

ig0774:
I'm posting this PR for comments.

The majority of issues we get seem to be caused by configuration issues, mostly stemming from Sublime running in an environment that users don't expect it to be. This is generally true of OS X as is well-known and at least possible on Ubuntu  (see #655, where the issue is, in essence, that Unity doesn't inherit it's environment from the login shell). In addition, because of the various places configurations can be made (either the `LaTeXTools.sublime-settings` or the project file) it could conceivably be quite hard to track-down which configuration setting is used. In short, we have (and provide) little visibility into the system so that users can understand what is going on and so we can help them troubleshoot their issues.

So I've created a new command which does some basic sanity checks and displays the information in a (hopefully) clear format.

Below is what it currently looks like:

Screen shot showing the output of the command

There are some differences depending on configuration, i.e., if a program can't be found or executed the status changes to 'missing'. In addition, on MiKTeX, it checks for `texify` instead of `latexmk` and the various environment variables `texify` uses (`BIBTEX`, `LATEX`, etc.).

Short TODO list:
- [x] Verify it works on:
  - [x] Windows 
    - [x] ST2
    - [x] ST3
  - [x] Linux
    - [x] ST2
    - [x] ST3
  - [x] OS X
    - [x] ST2
    - [x] ST3
- [x] Verify changes to `get_settings()` don't break anything
- [x] Add something to kill potentially hanging processes
- [x] Add checks for engine, tex options, tex root if launched from a tex file
- [x] Update everything for new (v3.7.x) code

Are there any other keys bits of information that are missing or would be useful to have? Could anything be done to make the information displayed clearer?

(I've thought about trying to get the actual commands the builder will run, but that turns out to be a major headache to do without actually running the commands or refactoring the builders which may be a possibility down the line.)

### DIFF
--- a/01_reload_submodules.py
+++ b/01_reload_submodules.py
@@ -22,7 +22,6 @@ LOAD_ORDER = [
     'external.latex_chars',
 
     'latextools_plugin_internal',
-    'latextools_plugin',
 
     # reloaded here so that makePDF imports the current version
     'parseTeXlog',
@@ -49,7 +48,9 @@ LOAD_ORDER = [
     'latextools_utils.analysis',
     'latextools_utils.ana_utils',
     'latextools_utils.bibcache',
-    'latextools_utils.output_directory'
+    'latextools_utils.output_directory',
+
+    'latextools_plugin'
 ]
 
 for suffix in LOAD_ORDER:
--- a/01_reload_submodules.py
+++ b/01_reload_submodules.py
@@ -48,7 +48,6 @@ LOAD_ORDER = [
     # depend on any previous
     'latextools_utils.analysis',
     'latextools_utils.ana_utils',
-    'latextools_utils.output_directory',
     'latextools_utils.bibcache',
     'latextools_utils.output_directory'
 ]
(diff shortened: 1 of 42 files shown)


r-stein:
I have tried it on my windows machine

- `ST3` works and displays the correct results, however it sometimes insert `CR` characters at the end of the line (I think ST buffers always use `LF` only, even with `CRLF` files)
- `ST2` also displays the result and does not have the `CR` problem, but does not support the :white_check_mark: and :x: characters

In addition `biber` is listed as required, although it must be explicitly enabled with texify. Hence I would list it as optional.

ig0774:
@r-stein: Thanks. I've fixed the `CRLF` thing and removed the required column altogether. May have to do the same with the dingbats as I can't get them to work on Windows 7 (I think 8 & 10 have some extra handling for that range of Unicode).

ig0774:
A version of this change has been merged.

PR #667, opened 2016-03-13
Add pdfviewer configuration for OSX

larryxiao:


r-stein:
I dont think it is useful to add this to `viewPDF.py`, but not to `jumpToPDF.py`. That would be more complicated, because it requires to implement forward sync (Jump to the position in the pdf).

larryxiao:
I agree. This is only a temporary solution to allow user to use custom apps (such as Preview) for viewing pdf.

larryxiao:
I just found #639! :)

PR #672, opened 2016-03-25
Autodetect Sumatra executable

r-stein:
This PR adds the ability to guess the executable of Sumatra by testing whether it is in the system path and checking the default installation paths.

Therefore the default settings for sumatra are changed to `null` (`None`) to trigger the detection. If the empty string `""` is used. it will assume Sumatra to be in the path (this is the current setup).

Additional default installation paths can be added easily.

### DIFF
--- a/LaTeXTools.sublime-settings
+++ b/LaTeXTools.sublime-settings
@@ -110,8 +110,10 @@
 		"texpath" : "",
 		// TeX distro: "miktex" or "texlive"
 		"distro" : "miktex",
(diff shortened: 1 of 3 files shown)


r-stein:
I just saw this is part of https://github.com/SublimeText/LaTeXTools/pull/639 and hence I close this PR.

ig0774:
Actually, thanks for doing this. This PR gave me some good pointers on things I wasn't doing correctly in #639.

PR #673, opened 2016-03-28
Adds a standard way to run external commands.

ig0774:
This standardises the various ways that we call subprocess.Popen(). My main future intention here is to be able to integrate [shellenv](https://github.com/codexns/shellenv), which should enable us to fully resolve some of the PATH issues, so that things get closer to "just working."

In addition, it ensures some consistency in:
1. Ensuring the PATH is set to `texpath`
2. Ensuring that we do not open a console window on Windows
3. It tries to log any commands executed in as sane a format as possible

I've also added versions of `check_call()` and `check_output()`.

Lastly, to ensure `get_texroot()` could be used wherever, I added a function (`run_on_main_thread`) which ensures that a function is called on the main ST thread on ST2 and returns the result. This enables the function to use the Sublime API, while the main processing can be done on a separate thread. As part of this, I cleaned up the latex_gen_pkg_cache command so it will always be run in a thread. More generally, this will make it possible to write sane threaded code to enable some form of background processing for long-running tasks.

### DIFF
--- a/01_reload_submodules.py
+++ b/01_reload_submodules.py
@@ -43,6 +43,7 @@ LOAD_ORDER = [
     'latextools_utils.sublime_utils',
     'latextools_utils.cache',
     'latextools_utils.quickpanel',
+    'latextools_utils.external_command',
 
     # depend on any previous
     'latextools_utils.analysis',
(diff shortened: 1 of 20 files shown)


r-stein:
@ig0774 It would be nice if you could resolve the merge conflicts, since the master also contains semantical changes (usage of the return code).

ig0774:
A version of this change has been merged.

Issue #559, opened 2015-09-27
How to configure the files generated after built?


Voyga:
Things like `.aux`, `.log`, `.synctex.gz` will be generated after built.
How do I configure the output files if I only want the `.pdf`?

ig0774:
You can't really configure LaTeX not to generate `.aux` and `.log` files. `.aux` is essential to LaTeX's operation and `.log` is essential to LaTeXTools understanding what happened (the messages about errors and warnings are generated from reading the log). So configuring a build to not generate those files isn't really an option. You can, however, run the `LaTeXTools: Delete Temporary Files` command available either on the command palette (`ctrl + shift + p` on most systems, `⌘ + shift + p` on Mac) or using the shortcut (`ctrl + l, backspace` or `⌘ + l, delete`), which will clear all those files.

`.synctex.gz` is a bit different because it isn't actually essential to the LaTeX build process. If you are using the traditional builder (and only the traditional builder, since the commands for the simple builder are hard-coded) you can change the `command` option in the `builder_settings` section of your `LaTeXTools.sublime-settings` file. The default is something like this: `"latexmk -cd -e pdflatex = '%E -interaction=nonstopmode -synctex=1 %S %0' -f -pdf"`. Just change that to `"latexmk -cd -e pdflatex = '%E -interaction=nonstopmode %S %0' -f -pdf"`. Note, however, that doing so will break the jump to pdf functionality, and I don't know what the outcome will be (i.e. if it will work at all).

Your best option, however, is just to run the `LaTeXTools: Delete Temporary Files` command after building.

Voyga:
Thanks for explaining!

Yes I think run `LaTeXTools: Delete Temporary Files` is my best option.

jirkadus:
Any chance to at least set the path for all, but pdf files to be stored in separate folder? Any ideas? Miktex has a -aux-directory=DIR option, is it possible to use it in builder settings within the LaTeX Tools?

ig0774:
Support for `-aux-directory` and `-output-directory` has been frequently requested and is somewhere on the TODO list (e.g. #98, #161, #163). Part of the problem is supporting things in a cross-platform way (e.g., TeXLive does not support `-aux-directory`; MiKTeX and TeXLive consequently have different interpretations of what `-output-directory` means) and part of the problem is just ensuring that we understand these implications and implement things in a sane way.

@Voyga: Are you happy to close this issue?

ig0774:
[...]

Issue #383, opened 2014-04-23
How to compile auxiliary files to a different folder on TexLive/Mac?


nunoh:
With so many updates to the build files in LatexTools and LaTeX compilers themselves, I admit I am confused about what and where I should change in order to be able to have something like automatically place all the temporary build files such as the ``.aux, .log, .bbl``, etc.. in something like a ``tmp`` folder.

Naturally, browsing the internet I came across some solutions that have worked for others, but not for me (yet) though. They are the optional flags of: ``-aux-directory=/some/temp/dir <additional options>`` in this SO thread (http://stackoverflow.com/questions/3745908/i-dont-want-the-aux-log-and-synctex-gz-files-when-using-pdflatex).

Has anyone been able to achieve something like that and would care to share or simply knows how to do it and/or can guide me doing that? That would be great!

I am using TexLive 2013 and on a Mac. 

Many thanks.

ig0774:
Unfortunately, this isn't really possible without changes to TeXLive. the `-aux-directory` flag is only implemented in the MiKTeX version of `pdflatex` and friends, so its limited to MiKTeX and hence to Windows. There is an `-output-directory` option, but that will move the final output, e.g., `.pdf` or `.dvi` file as well, which may be what you want to achieve... 

Obviously, it's possible to right a script to move the temporary build files, but then you may as well have deleted them, as they won't be picked up on the next build. (Which you can do, incidentally, but pressing `⌘+L, delete`).

Although it's possible to change the command used by the traditional builder to set either the `-aux-directory` or `-output-directory` option, I'd note that using either of those will cause LaTeXTools not to function properly (using `-aux-directory` it will not find the log file; using `-output-directory` it will not find the pdf either).

Perhaps it would meet your needs if a flag was added to enable the deletion of the generated files after build? That could be implemented fairly simply.

msiniscalchi:
Duplicate of #161 (and something that **eventually** we'll tackle once and for all, but not now). Closing.

Issue #169, opened 2013-04-17
Log files in a different output directory


subangstrom:
When using the $out_dir variable set in .latexmkrc  (for example: $out_dir = 'build';), the log file is not parsed by latextools. A user setting read by makePDF might be a quick fix

ig0774:
Support for `--output-directory` and `--aux-directory` is added in v3.8.0. Note, however, that this support will not work if you specify the `$out_dir` in `.latexmkrc`. LaTeXTools needs to know about the output directory in order to find log files, the pdf, etc.

vins31:
It does not work for me.
Instead of the `$out_dir` variable I use the `--output-directory` option, as indicated by @ig0774 .
Here is the content of my `latexmkrc` file

```latexmk
$pdf_mode = "1";
$pdflatex = "pdflatex -synctex=1 --output-directory build";
@default_files = ('test.tex');
```

I get the error `Could not read log file test.log`, using LatexTools v3.14.1 on OS X with Sublime Text 3.

ig0774:
Sorry if this wasn’t clear, but LaTeXTools doesn’t read `.latexmkrc` files. In other words, you have to add the `output_directory` setting.

Issue #163, opened 2013-04-01
OSX -output-directory option mimics -aux-directory option


lawlist:
https://github.com/lawlist/LaTexTools-Unofficial-Modification

LaTexTools-Unofficial-Modification
==================================

As to an OSX operating system, I have created an unofficial modification of the existing LaTexTools code to permit an -output-directory option to /tmp, and then copy the *.pdf back to the working directory.

[NOTE: The '/tmp' directory option is hard-coded into all four (4) of the files mentioned below; so anyone looking to change those portions would need to look for "## BEGIN / END modification by lawlist" within the code.]

Build now triggers an automatic save of the *.tex file, and the build window closes upon successful completion (or stays open if there is an error). The *.pdf that is used by Skim is the one in the /tmp directory, which works best for anyone looking to take advantage of the LaTexTools magic of jumpToPDF.py and viewPDF.py. This modification appears to handle the log issues (as to one *.tex file being built at a time), and a few other issues encountered relating to daily operations with the files below. Because my business only involves letters and legal pleadings (rarely with tables of contents and authorities), I have not yet had any need to modify files other than those below:

LaTeX.sublime-build

jumpToPDF.py

makePDF.py

viewPDF.py

I uploaded the original code before my modifications, so anyone who is interested can see the history of changes.

For anyone seeking the -aux-directory option with TexLive on OSX, please put in a feature request to the TexLive maintainers. Until the -aux-directory option is supported by TexLive, the best workarounds will probably be using the -output-directory option and copying the *.pdf to the working directory; or, simply cleaning up with the little -c option and a .latexmkrc file (defining the scope of the cleanup). The magic of jumpToPDF.py and viewPDF.py makes cleaning up the less preferred option, as the auxiliary files are needed for the magic to function correctly.

----

I created my own ST2 plugin that saves / builds / cleans (if build was successful), but it does not have the bells and whistles of LaTexTools. Things like forward / backwards sync would of course need the auxiliary files, and an automatic cleanup makes those options unavailable. It is only designed for building one (1) *.tex file in the open tab, and some of the output is piped into the python console. It is based upon the Bash concept of: latexmk file.tex && latexmk -c file.tex

https://github.com/lawlist/ST2-plugin-latexmk-save-build-clean

----

[...]

PR #674, opened 2016-04-03
Add support for --output-directory, --aux-directory, and --jobname

ig0774:
This adds the much-requested ability to use `--output-directory` and `--aux-directory` (cf. #559, #98, #161, #163, #383, #169). I've also added support for `--jobname`, as it pretty much has the same requirements and touches the same parts of the code.

All are controlled through either a setting or (yet another) `%!TEX` directive. The setting and directive name are `output_directory`, `aux_directory`, and `jobname` respectively. In addition, these changes will pick up any `--aux-directory=...`, `--output-directory=...`, or `--jobname=...` flags set via the TeX options support.

`output_directory` and `aux_directory` can be set to any of the following values:

* a path (relative to the main tex file, absolute, or, if specified in a project file, relative to the project file)
* `<<temp>>` a value which specifies to use a unique directory in the system temp path
* `<<cache>>` which specifies to use a unique directory in the ST cache directory or a subfolder of the User directory on ST2 (which doesn't have a cache folder)
* `<<project>>` which uses the same directory name as the other two but as a subfolder of the main file (I don't imagine this last option to be very useful to anyone; it was useful to build and test the feature).

`jobname` is just a string with no special interpolation.

In addition, I've added a setting `copy_output_on_build` which indicates whether to copy the PDF or any other files to the same folder as the main tex file. It can be set to:

* `true` (default) indicating to copy the PDF (only)
* `false` indicating that it should do nothing
* a list of extensions (e.g. `[".synctex.gz", ".pdf"]`, indicating that those files with the same basename as the main tex file and the given extensions should be copied.

When an auxiliary directory or output directory is used, it will be searched first for the log file. The output directory will be searched for the pdf file. If these files cannot be found it will fallback to the usual behaviour, i.e., a log or pdf file in the main directory. In addition, the behaviour of the `delete_temp_files` command is changed to delete the entire directory tree of the output directory and the auxiliary directory(+).

The hard part about adding this feature turned out not to be updating the various bits of code to be aware of the auxiliary directory, output directory or jobname. Rather, it was ensuring that the builders work properly. `latexmk`, whether on MiKTeX or TeXLive supports everything beautifully and simply. If you are intending to use the output directory feature, I'd highly recommend using `latexmk`. It turns out to be *impossible* to support this feature using `texify` because `texify` will fail with a fatal error if you attempt to pass `--output-directory`, `--aux-directory`, or `--jobname`, understandably(++). Thus, if you are using `texify`, none of these settings will not work (sorry; there's nothing I can do).

Since I figured it was better to *something* capable of supporting MiKTeX out of the box, I've added a new builder called `basic`(+++). In essence, it does the same thing as the `simple` builder with a few extra features:

1. It supports most of the same options / directives as the `traditional` builder, namely the `options` setting, the `program` setting, and (obviously) the `aux_directory`, `output_directory` and `jobname` settings as well as the corresponding directives.
1. It supports biblatex, including using biber, bibtex, or bibtex8 as engines (again, a commonly requested feature).
1. It supports running bibtex using `aux_directory` or `output_directory`, using a trick I borrowed from `latexmk`.

I've tried it out on some simple test cases and it works well enough though note that unlike `texify` it has no support for `makeindex` (patches are welcome).

I also modified the `script` builder to support the output directory via new `$aux_directory`, `$output_directory` and `$jobname` variables. However, in order to make it viable to use the script builder to run `bibtex`, I had to change the script builder so that it uses the shell (or cmd.exe) in order to allow basic shell commands. It should continue to function as it did before, but the added shell interpolation may affect some script commands people were using. In addition, `cmd.exe` doesn't work with a UNC path as the current working directory. If you're using the script builder for a project opened via a UNC path (not a mapped drive), you'll have to ensure all your paths are specified as absolute paths. Sorry about that.

(+) Strictly speaking, on Windows it does not remove the entire directory tree; it removes all the contents of the tree. The reason is that Sumatra has a directory change notifier attached to the directory of the file containing the PDF its viewing. The notifier does not receive messages when the directory it is watching is deleted (or if it does, it ignores them), which means that the directory becomes unusable until the PDF is closed in Sumatra. This seemed to be the path of least resistance.

(++) I say "understandably" because TeXify faces much the same problems LaTeXTools does: how to you read a log file when its name and location can be changed?

(+++) I didn't just try to add this to the `simple` builder as I agree with @msiniscalchi's reasons for keeping *that* builder as simple as possible. Note that, for this reason, the `output_directory` and `aux_directory` settings will be ignored if you are using the `simple` builder.

### DIFF
--- a/00_reload_submodules.py
+++ b/00_reload_submodules.py
@@ -26,12 +26,14 @@ LOAD_ORDER = [
     'latextools_utils.tex_directives',
 
     # depend on previous only
+    'latextools_utils.distro_utils',
     'latextools_utils.is_tex_file',
     'latextools_utils.sublime_utils',
     'latextools_utils.cache',
 
     # depend on any previous
     'latextools_utils.analysis',
+    'latextools_utils.output_directory',
 
     'latextools_plugin_internal',
 
--- a/01_temp_file_cleanup.py
+++ b/01_temp_file_cleanup.py
@@ -1,0 +1,69 @@
+import json
+import os
+import sublime
+import shutil
+import tempfile
+
+if sublime.version() < '3000':
+    def get_cache_directory():
+        return os.path.join(
+            sublime.packages_path(),
+            'User',
+            '.lt_cache'
+        )
+
+    strbase = basestring
+else:
+    def get_cache_directory():
+        return os.path.join(
+            sublime.cache_path(),
+            'LaTeXTools'
+        )
+
+    strbase = str
+
+
+# unfortunately, there is no reliable way to do clean-up on exit in ST
+# see https://github.com/SublimeTextIssues/Core/issues/10
+# here we cleanup any directories listed in the temporary_output_dirs
+# file as having been previously created by the plugin
+
+def plugin_loaded():
+    temporary_output_dirs = os.path.join(
+        get_cache_directory(),
+        'temporary_output_dirs'
+    )
+
+    if os.path.exists(temporary_output_dirs):
+        with open(temporary_output_dirs, 'r') as f:
+            data = json.load(f)
+
+        tempdir = tempfile.gettempdir()
+
+        try:
+            for directory in data['directories']:
+                # shutil.rmtree is a rather blunt tool, so here we try to
+                # ensure we are only deleting legitimate temporary files
+                if (
+                    directory is None or
+                    not isinstance(directory, strbase) or
+                    not directory.startswith(tempdir)
+                ):
+                    continue
+
+                try:
+                    shutil.rmtree(directory)
+                except OSError:
+                    pass
+                else:
+                    print(u'Deleted old temp directory ' + directory)
+        except KeyError:
+            pass
+
+        try:
+            os.remove(temporary_output_dirs)
+        except OSError:
+            pass
+
+if sublime.version() < '3000':
+    plugin_loaded()
--- a/LaTeXTools.sublime-settings
+++ b/LaTeXTools.sublime-settings
@@ -148,6 +148,96 @@
 	},
 
 // ------------------------------------------------------------------
+// Output Directory settings settings
+// ------------------------------------------------------------------
+
+	// OPTION: "aux_directory"
+	// Specifies the auxiliary directory
+	// Possible values:
+	//
+	// ""				the default; does not use any auxiliary directory
+	//
+	// path				the path to the auxiliary directory; if this is
+	//					not an absolute path it is interpreted as a
+	//					path relative to the main tex file
+	//
+	// "<<temp>>"		the auxiliary directory will be a temporary
+	//					directory generated in as secure a manner as
+	//					possible; note that this temporary directory
+	//					is only valid until ST is restarted and will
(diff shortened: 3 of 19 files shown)


arahlin:
This PR looks great!  I hacked my way around the build directory issue with a custom builder and using a latexmkrc file, but this is much cleaner.

Semi-random request: I think it would be useful to have `output_directory` and `aux_directory` that are set from the project settings file be (optionally) relative to the project root, rather than the default relative-to-tex-root.  For example, for a large project I tend to break the source into chapter directories and work on one chapter at a time, with all output files going into a single build directory shared by all of them.

ig0774:
@arahlin: I see your point. What it occurs to me to do is to repurpose `<<project>>` so that it gets interpreted as the path to the project file. That way the path could be specified as `<<project>>/output` (or whatever).

arahlin:
Sure, that seems reasonable.  I was thinking something a bit different: a relative path specified in the project settings file gets interpreted differently than a relative path specified as a TeX directive or in the user's LaTeXTools.sublime-settings file.  This seems more intuitive to me, but maybe just complicating things?

ig0774:
@arahlin: Ok, I've implemented things as you suggested. It makes the code slightly more complex, but I can see how its a more intuitive way for things to be handled. Thanks for the suggestion!

stephenlevitt:
A suggestion for this new feature: Would it be possible to add a setting for hiding the ".synctex.gz" file that is generated, using the file's attribute? This is desirable because then your working directory appears to only contain your tex source and its pdf output. The synctex file is still present, albeit hidden, so LaTeX-PDF synchronisation can still occur.

ig0774:
@stephenlevitt: That's worth considering, but probably as a separate feature, since it's probably useful outside the context of using `--output-directory` and `--aux-directory` as well.

stephenlevitt:
@ig0774 Sounds good. Thanks for your work on this project!

aadityadar:
I am not sure if this is the right place to ask this question but I wanted to ask if setting --output-directory and --aux-directory works when using MiKTex on Windows 10. When I define the paths to these directories in latextools.sublime-settings and compile my tex file, the auxiliary and output files are not generated in their respective folders. Is this because of what is documented in the readme file: "If you are using texify (default when using MiKTeX) or the simple builder, setting these flags through any method will be ignored" or am I making a mistake somewhere? Thank you.

ig0774:
What @r-stein said in #98 applies here too. You need to use the `basic` builder on MiKTeX or install `latexmk` and tell LaTeXTools that the distribution is `texlive`. The rationale is explain in [my first comment](https://github.com/SublimeText/LaTeXTools/pull/674#issue-145515843) on this PR.

aadityadar:
Thank you. My problem has been resolved. I request you to add the description of "basic" to the comments in the settings file as well. Currently, it gives options for default, traditional, scrip and custom.

ig0774:
Ooooh... Good catch! Btw, I realised your problem had been solved, I just wanted to add something in case someone comes across this later.

Issue #679, opened 2016-04-07
file helper for \inputminted (code files)


JimAmuro:
Hi,
I love the Fill Helper function very much. Now I want to use `\inputminted` to insert code file, is there any way to modify the package to make it also autofill the file name for code files? Thanks!

r-stein:
This is currently not possible, but I think this is a nice suggestion for future releases.

Adding it to the dropdown menu, which appears when pressing `{` should not be to complicated. However using it in the autocompletion would be harder, but still possible.

JimAmuro:
The former thing is what I mean. I just want to add it to the dropdown menu, i.e. a code file version for `\includegraphics{` . I think it is not so difficult because the treatment is the same, maybe I just need to add `\inputminted` into triggers and a filter for code files, but I don't know which file to modify...

r-stein:
If you want to temporary modify it on your own:
(Assuming you have some basic python knowledge)
You can open the file by pasting the following in the ST console `` ctrl+` ``:

``` py
window.open_file(sublime.packages_path() + "/LaTeXTools/latex_input_completions.py")
```

Afterwards add your regex in reversed order in line `37`, add a new "filter" in line `95` afterwards add the code in the `elif` block in line `133` similar to the svg part.

However with more and more commands this requires some rework and your changes are only temporary and might be overwritten with the next update.

JimAmuro:
Thank you very much.
I know it is just a workaround. It would be better if it could be added in the settings.

ig0774:
If you come up with something, feel free to [open a pull request](https://help.github.com/articles/creating-a-pull-request/). We're always happy to incorporate changes.

In terms of the settings, how would you envision that working?

r-stein:
@ig0774 I have a very promising approach [here](https://github.com/r-stein/LaTeXTools/tree/fillall_helper_setting_entries), which should also be able to solve https://github.com/SublimeText/LaTeXTools/issues/665

It builds the `TEX_INPUT_FILE_REGEX` and so on from dicts, which can also come from the user settings.

A setting entry could be:
```js
"fillall_helper_entries": [
	{
		"regex": "(?:\\][^{}\\[\\]]*\\[)?gvsedulcni\\\\",
		"extensions": ["svg"],
		"strip_extensions": [".svg"]
	},
],
```

ig0774:
@r-stein: That looks really good!

JimAmuro:
Thank you very much for your work on this! 
[...]

Issue #665, opened 2016-03-08
completion for includegraphics: other directory and macros ? 


pierremomo:
Hello,

I just discovered the new feature: completion for input/includegraphics .. it's terrific ! I have some questions: does it work only for images in the current directory ? or is there a way to specify some other subdirectory for example ? 

Moreover, I usually work with some macro, such as
 \newcommand{\image}[2]{\includegraphics[width=#1\linewidth]{./media/#2}}

is there a way to make the completion work also for this macro \image ? 

thanks a lot for your help and congrats again for your incredible work with this package...

Pierre

r-stein:
To address your questions and try to provide workarounds:

1. Currently it is not possible to add additional directories for image (or tex or bib) files. However all subdirectories of the tex root are scanned and you might create the desired behavior using symbolic links. This is not perfect, but might be an okayish workaround.
2. The completion selection is currently hard coded in the source code. Hence there is no option in LaTeXTools to use our macro. In addition other commands like "open included graphics" (`C-l,C-o`) also rely on the command being named `includegraphics`. In general I would say supporting newcommand definitions is hard to establish. However you could add your own keybinding to change the text and toggle between `image` and `includegraphics`:

  ``` js
  {
      "keys": ["{"],
      "command": "chain",
      "args": {
          "commands": [
              ["delete_word", {"forward": false}],
              ["insert", {"characters": "includegraphics"}],
              ["latex_fill_input", {"insert_char": "{"}]
          ]
      },
      "context":
      [
          {"key": "selector", "operator": "equal", "operand": "text.tex.latex"},
          {"key": "selection_empty", "operator": "equal", "operand": true, "match_all": true},
          {"key": "preceding_text", "operator": "regex_contains", "operand": "\\\\image$", "match_all": true}
      ]
  },
  {
      "keys": ["}"],
      "command": "chain",
      "args": {
          "commands": [
              ["move_to", {"forward": true, "to": "brackets"}],
              ["delete_word", {"forward": false}],
              ["insert", {"characters": "image"}],
              ["move_to", {"forward": true, "to": "brackets"}]
          ]
      },
      "context":
      [
          {"key": "selector", "operator": "equal", "operand": "text.tex.latex"},
          {"key": "selection_empty", "operator": "equal", "operand": true, "match_all": true},
          {"key": "preceding_text", "operator": "regex_contains", "operand": "\\\\includegraphics{[^}]*}$", "match_all": true}
      ]
  },
  ```
[...]

PR #681, opened 2016-04-09
Rework of input completions

r-stein:
This PR reworks the input-completion functionality. The main purpose was to add flexibility and user definable completions. It has been shortly discussed in https://github.com/SublimeText/LaTeXTools/issues/679 and will also solve https://github.com/SublimeText/LaTeXTools/issues/665.

A potential settings entry could be:

``` js
"fillall_helper_entries": [
	{
		"regex": "(?:\\][^{}\\[\\]]*\\[)?gvsedulcni\\\\",
		"extensions": ["svg"],
		"strip_extensions": [".svg"]
	},
],
```

Changes:

- completions (trigger regex, file extensions) are now defined as json compatible dicts and are also definable in the user settings
- a context has been added, such that is it possible to add completions without updating the keymap for `{`
- input completions can now optionally be triggered via `,` (e.g. for `\addbibresource`). (the key for this is `"comma_separated": True`)

TODO:
- ~~README entry~~
- [x] entry and description in default settings
- ~~additional options, (e.g. change the search folder)~~


JimAmuro:
I'm sorry but still no luck for `\includesvg` and `\lstinput`...
I changed all the files here. No error message, it just writes `{}` in the file after I press `{`.

r-stein:
@JimAmuro you could try to download or clone (`git clone https://github.com/r-stein/LaTeXTools -b fillall_helper_setting_entries`) the whole folder

LaTeXTools.sublime-settings
```diff
@@ -282,6 +282,40 @@
 		]
 	},
 
+
+// ------------------------------------------------------------------
+// Additional fill all helper settings (for experts)
+// ------------------------------------------------------------------
+	// entries for the fill all helper
```
ig0774: It'd be good to have some documentation on the format of entries expected... Not everyone may be willing to read the Python code.

JimAmuro:
@r-stein I tried clone your repo today. Still no luck for entries in "fillall_helper_entries"... Windows 10. ST3. Deleted my user settings for LaTeXTools.

r-stein:
@JimAmuro thats weird, you could look at the console `` ctrl+` `` output, when saving `latex_input_completions.py` and triggering `\includesvg{`

JimAmuro:
@r-stein I have tried that, no error information.

ig0774:
@r-stein: It might be a good idea to segregate out the fill-all helpers into their own file, separate from the main settings. At a technical level, there's no real justification for it, I just worry that this is going to make the already-intimidating settings file even worse.

r-stein:
@ig0774 I am also struggeling with the complexity of the settings file. My idea would be to divide it into an advanced/expert settings file and normal settings file. My suggestion would be to add an additional settings file (`LaTeXToolsAdvanced.sublime-settings`), which is also accessable via the menu. If someone wants to overwrite an advances setting, he just copies the entry to his normal settings and changes it.
Hence the `get_settings` function would load `view.settings > LaTeXTools.sublime-settings > LaTeXToolsAdvanced.sublime-settings`.
Using this, one could move the cache or `open_image_command` setting to the advanced settings.

ig0774:
@r-stein I suppose your proposal makes the most sense in terms of the simplest migration pathway. I'm a little hesitant about making `get_settings()` any more complex, but I suppose as long as we have a way to determine what the relevant setting is (which is part of the intention of #656), everything is ok.

ig0774:
Shouldn't we also modify FillAll to leverage the `on_query_context` so it will be able to support dynamic entries as well as the static once?

r-stein:
The `on_query_context` event listener supports both static and dynamic entries.

ig0774:
@r-stein: Is this ready-to-go?

r-stein:
No, I still want to add some documentation. Features and addition arguments (e.g. changing the search folder) could be a future PR.

latextools_utils/is_tex_file.py
```diff
@@ -12,7 +12,6 @@
     strbase = str
 
 def get_tex_extensions():
-    view = sublime.active_window().active_view()
```
r-stein: I removed this, because it raises in error when called inside `plugin_loaded` in ST2 and is not used anyway.

ig0774:
Documentation-wise, this might do well as a [wiki page](https://github.com/SublimeText/LaTeXTools/wiki). I'm not sure anyone actually reads them, but at least we'd have something.

r-stein:
Agree that would be better.

PR #683, opened 2016-04-19
Jump to Anywhere

r-stein:
The "jump to anywhere" is a smart context dependent mouse driven IDE-like jump. The user just presses `alt-leftclick` and a jump will be executed depending on the command on which the user clicks. Supported jump types are:

- Jump to referenced labels (e.g. `\ref`)
- Jump to citation entries in bibliography files (e.g. `\cite`)
- Open included files (e.g. `\input` or `\include`)
- Open included graphics with a specified program (e.g. `\includegraphics`)
- Open the documentation of used packages (e.g. `\usepackage`)
- Jump to self-defined command definition, i.e. jump to the `\newcommand` in which the command was defined

However its easy to introduce additional jumps.

__Integration with SublimeCodeIntel__
Since the same mouse-binding is used by [SublimeCodeIntel](https://github.com/SublimeCodeIntel/SublimeCodeIntel) and ST has no context in mouse-binding, this package has special integration for SublimeCodeIntel. The mouse binding has a `fallback_command`, which will be executed outside of tex files. Because SublimeCodeIntel overwrites the mouse-binding *(`S` is lexically greater than `L`)*, there is a command to create a mouse-map in the user folder.

Demo:

jumpto_anywhere


excetara2:
I don't know if it is possible or already implemented as I haven't checked out this branch but if when you jump to a bib entry. If the file attribute is in the bib entry then you can click on that in the bib file and it will open the pdf as well. Just a suggestion if possible.

r-stein:
This PR is for jumps inside text documents and currently only supports tex commands. I don't know much about the syntex and jumping into the pdf behavior.

excetara2:
Yeah it looks like a really cool feature. I will give it a try soon. I just meant the `file` part of the bibentry below when it jumps to the bibtex files. But I guess that is outside the scope of the current PR.
```
@article{Devergie2011,
author = {Devergie, Aymeric and Grimault, Nicolas and Gaudrain, Etienne and Healy, Eric W and Berthommier, Fr{\'{e}}d{\'{e}}ric},
file = {:Users/excetara2/Dropbox/Research/Papers/Devergie et al.{\_}2011{\_}The effect of lip-reading on primary stream segregation.{\_}Journal of the Acoustical Society of America.pdf:pdf},
institution = {Centre de Recherche en Neurosciences de Lyon, UMR CNRS 5292 Université Lyon 1, 69366 Lyon Cedex 07, France.},
journal = {J. Acoust. Soc. Am.},
number = {1},
pages = {283--291},
pmid = {21786898},
title = {{The effect of lip-reading on primary stream segregation.}},
url = {http://link.aip.org/link/JASMAN/v130/i1/p283/s1{\&}Agg=doi},
volume = {130},
year = {2011}
}
```

r-stein:
Okay now I understand the suggestion. This and also opening the url seems to be a good idea.

excetara2:
Yes I agree and sometimes the `file` has other things inside like `:html` or `:ppt` (looked through and only found a couple cases out of 300 or 400 that weren't just a pdf. Pasted one below.). So these could likely be ignored or handled possibly. In case you implement I thought it might help to have a more full example.

```
file = {:Users/excetara2/Dropbox/Research/Papers/Shamma, Klein{\_}2000{\_}The case of the missing pitch templates how harmonic templates emerge in the early auditory system{\_}Journal of the Aco.ppt:ppt;:Users/excetara2/Dropbox/Research/Papers/Shamma, Klein{\_}2000{\_}The case of the missing pitch templates how harmonic templates emerge in the early auditory system{\_}Journal of the Aco.pdf:pdf},
```

r-stein:
@excetara2 I took a look and it seems feasible, althrough the drive (`C:\`) is stripped of on windows. I will include it in this or a follow up PR. Do you have a link where this is more specified?

excetara2:
@r-stein I actually can't find where any specification for this is listed. But it seems all the programs use it like Zotero and Mendeley from searching the internet. Mine was an example from Mendeley. I also just had a quick check and output from Papers just to see what the bibtex file looked like. I am not sue their is a standard format.
```
local-url = {file://localhost/Users/bigtyme/Dropbox/SyncedPrograms/Library.papers3/Files/51/51F9DDF9-6371-42C7-90E5-B99D9F77C043.pdf},
file = {{51F9DDF9-6371-42C7-90E5-B99D9F77C043.pdf:/Users/bigtyme/Dropbox/SyncedPrograms/Library.papers3/Files/51/51F9DDF9-6371-42C7-90E5-B99D9F77C043.pdf:application/pdf}},
uri = {\url{papers3://publication/uuid/4FDBB63F-127B-4C16-87F4-8A44978496BF}}
```

and another entry:

```
file = {{4B390CB0-6312-4EA8-80E9-C836F7F2BF38:/Users/bigtyme/Dropbox/SyncedPrograms/Library.papers3/Files/4B/4B390CB0-6312-4EA8-80E9-C836F7F2BF38:application/pdf}},
uri = {\url{papers3://publication/uuid/30B90AF9-258D-4758-A905-A86FD6C728FD}}
```

excetara2:
I think zoteros is the same as Mendeley from what I can see.

r-stein:
It is also used by jabref. I think the easiest way is just to split the entry by `:`, unescape special chars, and search for file names.

Default (OSX).sublime-mousemap
```diff
@@ -0,0 +1,12 @@
+[
+    {
+        "button": "button1", 
+        "modifiers": ["ctrl"],
```
ig0774: This should probably be `["alt"]`. `ctrl+button1` is used for right clicks on some older Macs (where there is usually only one button anyway).

r-stein: I used the same mouse bindings as SublimeCodeIntel ([osx mousemap](https://github.com/SublimeCodeIntel/SublimeCodeIntel/blob/master/Default%20%28OSX%29.sublime-mousemap)), because it is a very common package (TOP 25) and this should feel similar. However I have never used osx and don't know how the buttons are used and which buttons are available.

ig0774: Ah, yes. I see the point. I use the mouse so infrequently in ST I forgot about the various bindings. Still, something like `["ctrl", "super"]` while a bit less convenient at least doesn't interfere with the OS or any existing bindings. I'd also note that for Linux SublimeCodeIntel [uses `"super"`](https://github.com/SublimeCodeIntel/SublimeCodeIntel/blob/master/Default%20(Linux).sublime-mousemap), though I'm not sure exactly why.

excetara2:
I've finally had a bit of time to use this and opening images on OSX doesn't work for me. Using `ctrl+l ctrl+o` works fine to open but the open image command set to open in preview as below doesn't do anything. All other commands seem to work great. Maybe you can add `\includeonly` to the `\include` and `\subfile` to open these included files as well.

```
	"open_image_command": {
		"osx" : [
			{
				// open pdf and eps documents with skim
				"extension": ["pdf", "eps"],
				"command": "open -a Preview.app"
			}
			// uncomment these lines to open all other images with the default programm
			,
			{
				"command": "open"
			}
		],
```

excetara2:
okay here is the issue when using subfloats it doesn't work if the includegraphics is on the same line as the subloat command like below. If you hit enter after the \includegraphics then it works fine and opens. `Ctrl+l ctrl+o` works fine in this instance:

```
 \begin{figure}[ht!bp]
      \centering
      \subfloat[Piano]{\includegraphics[width=0.42\linewidth]{bionic/figures/bio_correct_recog_22_piano_1ms.pdf} %l b r t
      \label{fig:bio:piano_22_chan}}  
\end{figure}
```

PR #685, opened 2016-04-19
Generate PDF README

ig0774:
Probably most helpful to [look at this in action](https://github.com/ig0774/LaTeXTools/blob/revamp_readme/README.markdown).

This is an attempt to reorganize the README, which has been getting a bit scattered. I've also taken the opportunity to add some new documentation, remove out-of-date documentation, improve the documentation on some features, add some internal links (which hopefully improve navigation) and move everything to a unified hierarchy of section headings. 

This PR now uses [pandoc](http://pandoc.org/) and [minted](http://www.ctan.org/pkg/minted) to generate a PDF version of the README to be distributed with the package and made available via the `README` menu option.


excetara2:
I do a lot of converting pandoc documents so you just need to run: `pandoc README -f markdown_github --latex-engine=xelatex -o README.pdf`. Did a quick check and you will just have to fix the tables so they are aligned more properly in the markdown which is quite easy. If you search for `pdf_file`, I fixed that table to see what the others should look like:
[temp.pdf](https://github.com/SublimeText/LaTeXTools/files/243114/temp.pdf)

ig0774:
Does that work with Github's markdown as well? They seemed to me to require different formatting. I've been running it with `--columns=100` which seemed to fix the tables if reflow the text. My other concern was getting a TOC in a sensible place.

excetara2:
Yes that's what the `-f markdown_github` command tells pandoc that the input is GitHub flavored markdown. Yes I forgot about the `--columns=100` command as I always align my tables using `AlignTab: Table Mode` so usually fixes alignment issues with converting from different formats in pandoc.

Also I like the `-N` option to add numbered sections for a pdf but I guess that's personal preference. The only way I know to get the TOC somewhere else then the top is to modify the default template `--template` and pass in your own design that puts the table of contents elsewhere.

ig0774:
Thanks @excetara2. Those were some helpful tips. I got it to look a bit nicer. I had to actually write a [filter](http://pandoc.org/scripting.html) to get the TOC in the right place. On numbered sections, I'm fine with either having them or leaving them out, but I kind of incline toward leaving them out to match the README file.

In case anyone is wondering, [this](https://github.com/SublimeText/LaTeXTools/files/244287/README.pdf) is what the PDF produced by the code in 38908a1 looks like. [This](https://github.com/ig0774/LaTeXTools/blob/revamp_readme/README.markdown) is the Github-rendered markdown.

excetara2:
Yeah either way I think is okay on the numbering but just maybe easier to follow what section you are in further down in the document when numbered with the PDF.

My only comment is that you don't want the code blocks syntax highlighted like github?? I think it makes it much easier to read especially python syntax blocks. If you look at the pdf I linked, you can see what the syntax highlighting would look like for the default using `pygments`. I guess you'd put something in the preamble to overwrite the default syntax highlighting for github_markdown mode. Anyways other than that looks great.

ig0774:
@excetara2: Numbering might be nice, it's just not something I've dug into yet. I'm also trying to keep things close to the way the README is rendered on github right now.

Re: Syntax highlighting: It's desirable, but I put it at a lower priority than getting the code blocks to not overrun the margins. That's been addressed and I've added in some code highlights using `minted` in the latest version. [Here's a PDF preview](https://github.com/SublimeText/LaTeXTools/files/256898/README.pdf).

excetara2:
@ig0774 Yeah that looks great now. I agree that not overrunning the margins is probably more important and I think the `minted` highlighting is enough to make it easy to read.

ig0774:
Closing in favour of #952.

Issue #680, opened 2016-04-09
Command to jump to pdf but with focus on pdf not changing the toggle


excetara2:
I know there is the toggle switch and I can use chain of command to toggle, then jump to pdf, and then toggle back but seeing if the LatexTools has anything built in for this.

Sometimes I want to jump and have focus on the pdf so just curious if there was a function built-in.

r-stein:
This is currently not supported, but it might be useful and easy to add this.

@ig0774 changing line `159` to `keep_focus = args.get('keep_focus', get_setting('keep_focus', True))` would enable the user to overwrite the setting by passing args and one could e.g. use the keybinding:

``` js
    {   "keys": ["ctrl+l","J"],
        "context":  [
            {"key": "selector", "operator": "equal", "operand": "text.tex.latex"}],
        "command": "jump_to_pdf", "args": {"from_keybinding": true, "keep_focus": false}
    },
```
to let the focus at the pdf viewer and keep the focus when building

excetara2:
Yeah I was going to keep the original `["ctrl+l", "j"]` as is and just add ["ctrl+l", "ctrl+j"] to send focus to the pdf for the times it is necessary.

r-stein:
@excetara2 you can already test the change just open the console `` ctrl+` `` and paste `window.open_file(sublime.packages_path() + "/LaTeXTools/jumptoPDF.py:159", sublime.ENCODED_POSITION)`. Afterwards apply the change.

Aside: I would recommend to use the `ctrl+l, shift+j` keybinding, because I am looking forward to use your keybinding in a [future PR](https://github.com/r-stein/LaTeXTools/tree/jumpto_anywhere#jump-to-anywhere)

excetara2:
Okay yeah it works great so thanks and I have set it to a different binding
as that looks like a very useful PR. Ended up setting it to `ctrl+l`, `\`
because that's used in sublime's file browser to open external file browser
or open an application with the file so used to that jumping out of Sublime.

I assume at a later PR you'll add in the above code?

ig0774:
Closing as this is now possible.

PR #700, opened 2016-04-21
Allow focus of pdf editor by passing args for using keybindings

excetara2:
Fixed and updated issue as discussed in using r-stein solution #680

### DIFF
--- a/Default (Linux).sublime-keymap
+++ b/Default (Linux).sublime-keymap
@@ -329,6 +329,11 @@ LaTeX Package keymap for Linux
 			{"key": "selector", "operator": "equal", "operand": "text.tex.latex"}],
 		"command": "clear_local_latex_cache"},
 
+	// ctrl+l,w to display word count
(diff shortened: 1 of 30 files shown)


r-stein:
I think the change should be done on [line `116`](https://github.com/excetara2/LaTeXTools/blob/master/jumpToPDF.py#L116) and `forward_sync` could also be passed as an argument. And it would be nice to write something in the [README](https://github.com/excetara2/LaTeXTools#forward-and-inverse-search), so people have the chance to know it without reading the python code.

*Side remark: If you push on your branch, the commits will be included in the pull request. Hence you don't have to create a new one if you made changes. In addition I would recommend to not use your masterbranch for future PRs.*

excetara2:
Okay I've updated it so both arguments can be passed and updated the README to state the toggles can be overwritten by the passed in arguments.

_Any future pull requests I'll make a new branch_

ig0774:
Thanks! I've integrated this fix into v3.7.10.

PR #706, opened 2016-04-28
Improved TeXSyntaxListener for ST3 and user defined syntax

r-stein:
This PR changes the syntax listener to

- support of the new ST3 `*.sublime-syntax` files
- use `score_selector` to detect whether any tex syntax is set. This is more resistant to user self-defined syntax files, but also to special syntax support like Beamer or TikZ.

this has been mentioned by @keith-hall in https://github.com/SublimeText/LaTeXTools/commit/1baf0efd52dc2dee30c17eb6b99120f2fdb88abe#commitcomment-17285547


ig0774:
Thanks for this!

PR #711, opened 2016-05-03
Self-defined commands completions

r-stein:
This PR adds completions for self-defined commands and environments.

The completions are added into the cwl completions and are only available if cwl completions are available.

Supported completions:

- environment in `\(re)newenvironment`
- commands in `\(re)newcommand`
- math operators from `\DeclareMathOperator` (only inside math)

Feel free to suggest more.

demo

### DIFF
--- a/latex_own_command_completions.py
+++ b/latex_own_command_completions.py
@@ -30,11 +30,22 @@ def get_own_command_completion(view):
     if not tex_root:
         return []
 
+    # use special handling (additional completions) for math mode
+    math_selector = (
+        "string.other.math.tex, "
+        "string.other.math.latex, "
+        "string.other.math.block.environment.latex"
+    )
+    is_math = bool(view.score_selector(view.sel()[0].b, math_selector))
+
     def make_completions():
         ana = analysis.get_analysis(tex_root)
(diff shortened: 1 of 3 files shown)


latex_own_command_completions.py
```diff
@@ -0,0 +1,93 @@
...
+        s += "{arg}" * arg_count
+        comp = parse_keyword(s)
+    except:  # no args
+        s = c.args + "{}"
+        comp = s
+    return (s + "\tself-defined", comp)
+
+
+# TODO duplicate to avoid cyclic imports
+def parse_keyword(keyword):
```
r-stein: I just copied this function, because importing it would result in a cyclic import. I could import it from here into the other file or move it to an additional file. Has someone a better idea?

ig0774: My immediate thought here is why not just move the own_environment code to latex_env_completions and the own_command code to latex_cwl_completions.

r-stein: Yes I could join it with latex_cwl_completions (the env completion is only the fillAllHelper for `{`), but I think it would decrease the readability.

PR #715, opened 2016-05-07
Improved robustness of author parsing for cite completion

r-stein:
This slightly increases the robustness and readability of author parsing in cite completions.

Before this PR a bug occured if an author is empty `A and  and C`, because of the check `a[-1] != '}'`. The fix is just filter out empty strings via `if len(a) > 0`.


ig0774:
Thanks! This looks great.

PR #723, opened 2016-05-13
Made LaTeXTools compatible with LaTeX-Plus syntax definition

mpvharmelen:
Specifically, followed the advice from [this proposal](https://github.com/sublimehq/Packages/issues/340)
about adding `meta.function.environment.math.latex` to the maths completions scope.

### DIFF
--- a/Default (Linux).sublime-keymap
+++ b/Default (Linux).sublime-keymap
@@ -223,6 +223,7 @@ LaTeX Package keymap for Linux
 			{"key": "selector", "operator": "not_equal", "operand": "string.other.math.tex"},
 			{"key": "selector", "operator": "not_equal", "operand": "string.other.math.latex"},
 			{"key": "selector", "operator": "not_equal", "operand": "string.other.math.block.tex"},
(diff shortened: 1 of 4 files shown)


ig0774:
Thanks for this. I'm going to hold off on pulling anything until the final format of the new syntax is worked out ( see SublimeHQ/Packages#370). Likely we'll have to make some more adjustments down the road.

r-stein:
The upcoming LaTeX syntax uses the scope `meta.environment.math.latex`. However `meta.function.environment.math.latex` is already used in the standard syntax for math environments `\begin{equation}...\end{equation}`. Hence it would be useful to add both scopes.

ig0774:
@r-stein: That's a good point. I'd definitely merge changes that are useful today.

r-stein:
I just took a look and the scope is used in the same range as `string.other.math.block.environment.latex`, hence we do not to specially add this. I will add the new scope `text.tex.latex meta.environment.math` on my own and close this PR.
Thanks for it anyway.

PR #726, opened 2016-05-13
Also react on subref

raimund-schluessler:
This PR makes LatexTools to also react on the ```\subref{}``` command and auto-fill the reference.


ig0774:
Thanks for this! However, it actually duplicates #719. I haven't merged #719 yet because I wanted to get it to support `\subref*` as well. Should be in the next minor release.

raimund-schluessler:
Ok. Sorry that I didn't look for pending PRs.

Issue #499, opened 2015-04-26
Quick panel \ref with files using the \import command from the import package


watkinrt:
The import LaTeX package allows you to use complicated file trees (which is particularly helpful when writing a book or thesis). In the current implementation of Quick panels with the \ref command, only \include and \input commands are searched for. I've come up with a solution to additionally search for \import commands and then search through their file tree appropriately. At line 81 of latex_ref_completions.py is the following code:

``` python
# search through input tex files recursively
    for f in re.findall(r'\\(?:input|include)\{([^\{\}]+)\}', src_content):
        find_labels_in_files(rootdir, f, labels)
```

To also search for import commands (which need to be parsed somewhat differently), you can add code like this

``` python
importLines = re.findall(r'\\(?:import)[*]?\{([^\{\}]+)\}\{([^\{\}]+)\}', src_content)
    for (base,f) in importLines:
        find_labels_in_files(rootdir+"/"+base, f, labels)
```

after the above code. I've implemented this and it works for the cases I'm interested in; however, I haven't check to see how robust it is.

schinknbrot:
Nice one!

Thank you a lot!

ig0774:
Fixed in v3.13.0

PR #727, opened 2016-05-13
Also parse files included by import package

raimund-schluessler:
This change allows LatexTools to also parse external files which were included using the import package.


raimund-schluessler:
I just saw that this was requested by a different user in https://github.com/SublimeText/LaTeXTools/issues/499 already.

r-stein:
Sorry for the late reply . I close this, because it is superseded by #990.

PR #732, opened 2016-05-15
Less Obtrusive Build Panel

ig0774:
This builds on #712 and is intended to run builds in a less obtrusive way. Among other things, it implements #638 while addressing [my concerns here](https://github.com/SublimeText/LaTeXTools/pull/638#issuecomment-169167969).

This (if it is merged) is a rather major change, so I'm attempting to solicit some feedback.

1. By default, the output panel is not displayed initially unless `hide_build_panel` is set to `never`. Instead, an indicator (borrowed from [PackageControl](https://packagecontrol.io/)) is used to indicate the build is in progress.
1. By default, the output panel appears if any errors or warnings occur including (and this is the crucial bit) any exceptions thrown at the various points we already have "COULD NOT COMPILE" messages. If `hide_build_panel` is set to `always`, of course, the panel will never be shown.
1. I've added a key-binding, `shift+escape` to bring up the LaTeXTools output panel (this is the reason this change is based off #712 which, for other reasons, uses a uniquely named output panel). I realise a key-binding outside of the `C-l` namespace is risky, but it doesn't seem to be used frequently (if at all) and it corresponds well to the `escape` default binding which will already close the build panel.

The defaults may not be to everyone's tastes, but hopefully its a reasonable compromise.

### DIFF
--- a/Default (Linux).sublime-keymap
+++ b/Default (Linux).sublime-keymap
@@ -223,9 +223,12 @@ LaTeX Package keymap for Linux
 			{"key": "selector", "operator": "equal", "operand": "text.tex.latex"}],
 		"command": "insert_snippet", "args": {"name":"Packages/LaTeXTools/Text monospace.sublime-snippet"}},
 
-
-
-
+	// Show panel
+	{ "keys": ["shift+escape"],
+		"context": [
+			{"key": "selector", "operator": "equal", "operand": "text.tex.latex"},
+			{"key": "panel_visible", "operator": "equal", "operand": false }],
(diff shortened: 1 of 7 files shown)


mpvharmelen:
I think this is very nice! My only concern is:
> this is the reason this change is based off #712 which, for other reasons, uses a uniquely named output panel

Does this break the built in `show_panel` command that shows the build results? If it does I would definitely not use a uniquely named output panel.

ig0774:
It doesn't break the `show_panel` command at all. It simply changes the argument passed from `output.exec` to `output.latextools`, which is probably what we should have done in the first place.

PR #759, opened 2016-06-29
Update bibtex parser to support cryptobib

ig0774:
* Eliminates wrong-headed support for \" escapes in quoted strings
 * Adds support for brackets in quoted strings
 * Removes special case for @string values

- [x] Test this in LaTeXTools

### DIFF
--- a/external/bibtex/lexer.py
+++ b/external/bibtex/lexer.py
@@ -199,6 +199,7 @@ class Lexer(object):
                     bracket_depth += 1
                     value.append(matched)
                 else:
+                    self.current_line += 1
                     # consume space after new line replacing with 1 space
                     match = SPACE.match(self.code, i - 1)
                     if match:
@@ -234,9 +235,15 @@ class Lexer(object):
 
                 if matched == '"':
                     break
-                elif matched == '\\"':
-                    value.append('"')
+                elif matched == '{':
+                    self.current_index = match.start()
+                    consumed = self.value_token()
+                    if consumed > 0:
+                        i += consumed - 1
+                        last_token = self.tokens.pop(-1)
+                        value.extend(['{', last_token[1], '}'])
                 else:
+                    self.current_line += 1
                     # consume space after new line replacing with 1 space
                     match = SPACE.match(self.code, i - 1)
                     if match:
@@ -334,6 +341,6 @@ NUMBER              = re.compile(r'\d+', re.UNICODE)
 KEY                 = re.compile(r'([^\W\d][^,\s=]*)\s*=\s*', re.UNICODE)
 
 # These are used internally by the more complex "tokens"
-NEXT_QUOTE_BREAK    = re.compile(r'(?:\\")|\n|"')
+NEXT_QUOTE_BREAK    = re.compile(r'\n|"|\{')
 NEXT_BRACKET_BREAK  = re.compile(r'\{|}|\n')
 SPACE               = re.compile(r'\s+', re.UNICODE)
--- a/external/bibtex/lexer.py
+++ b/external/bibtex/lexer.py
@@ -171,14 +171,9 @@ class Lexer(object):
 
         return len(match.group(0))
 
-    def value_token(self):
-        i = self.current_index
-        if self.code[i] != '{':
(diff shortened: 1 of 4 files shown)


tlepoint:
Thanks. It seems to work better. A couple of remarks:

- It's still very slow even when getting the cache :-/ The following items do not solve anything, but are suggestions:
    - what about using the fastest protocol available for pickle? `pickle.dump(obj, f, protocol=-1)` in `cache.py`
    - I think the verification of the `cached_time` should be done first, _before_ the `cached_entries` are  loaded rather than at the same time (thus when we modify a bib file, it starts regenerating the cache right away, and does not load the cached values)

- From some verbose I did, reading the cache takes about 4 seconds on my computer, but the whole process of displaying the panel is longer than that. I suspect that giving to the panel a huge array takes a very long time. Is there a way to optimize that?

r-stein:
Thanks for the suggestions

- using a better protocol seems to be a nice and easy idea. I am also thinking about a multi-level cache, which keeps frequently used cache-entries in the memory to avoid file access and picklings.
- I don't think moving the verification makes a real difference at the end, because I don't think bib files (and especially such big big files) are changed that often.
- the performance problem seems to result from the formating of the entries and not from the creation of the quick panel. Hence we can optimize (and maybe also cache) that.

tlepoint:
Thanks.

Just another remark/question: there is a huge difference between _traditional_ and _new_ when it's reading from the cache, I am not sure why for the moment. This makes _new_ way less usable; I would expect only the caching to be longer?

> - I don't think moving the verification makes a real difference at the end, because I don't think bib files (and especially such big big files) are changed that often.

That's true, but it could be as simple as providing an optional parameter (a date) to the cache read function, which makes sure that something cached before this date is no longer useful. I just wanted to emphasize that it seems to me that the modification could be small.

r-stein:
Taking a look at the resulting cache files the cache file for the `new` bibliography requires way more space (24 MB vs 8 MB).
It would be useful to reduce the size of the cache file, e.g. by filtering the bib entry keys to obtain only the keys, which are necessary to create the quickpanel. E.g. remove something like `booktitle` or `url` if not used in the format string.

ig0774:
Yes, or perhaps just caching the entries already formatted for the quick panel. There's a fairly significant cost to the formatting itself when a thousand or more entries are involved, which needs to be paid each time the panel is loaded. This has three other advantages: 1) we don't need to parse the format string itself, 2) we could still cache the bibliography and avoid needing to reparse the whole thing if a new field were added to the format string, and 3) we can get the same speed-up for both bibliography plugins.

A more IDE-like option might be to pre-compute the bibliography quick panel i.e. on a background thread when the file is open. This is probably a good idea for extremely large bibliographies that don't change all that frequently, but it may be overkill for smaller bibliographies, especially if they change relatively frequently.

r-stein:
I tried a proof-of-concept implementation to cache the formated strings for the bib files. The whole command with the `crypto.bib` file (38k entries) took less than 1 second (0.65s - 0.95s). Creating the quick panel takes around 0.5s.
Hence I this is a practical solution and I will prepare a PR for this.

We can still run the caching process in the background so the user will experience the IDE-like solution.

r-stein:
The new bibcache implementation is available [here](https://github.com/r-stein/LaTeXTools/tree/bibcache).

It adds a `bibcache` module, which can be used by bibliographies to improve the performance.
The `bibcache` automatically formates the autocomplete and cite panel entries. These formatted entries are stored with the special keys `"<panel_formatted>"` and `"<autocomplete_formatted>"`.
Each bibentry is converted into a dict, which only contains the completions and some important keys like "keyword". It seems to be way faster to cache a dict than a class.

Extending it this way result in a fast completion even for big files and preservers compatibility for all bibliographies.

ig0774:
One very small thing: in `newBibliography.py` you left in the line `cache_name = "bib_" + hashlib.md5(bibfname.encode("utf8")).hexdigest()`, which should be deleted.

tlepoint:
That's very fast once the cache is built! Many thanks, that is much better.

PR #761, opened 2016-06-30
Add ability to jump to the root from %!TEX root comments

ig0774:
Simple change to allow jumping to the root document from an included document if the `%!TEX root` comment is used.

May not be terribly useful, but I've found myself wanting this.

### DIFF
--- a/jumpto_tex_file.py
+++ b/jumpto_tex_file.py
@@ -12,10 +12,12 @@ try:
     _ST3 = True
     from .getTeXRoot import get_tex_root
     from .latextools_utils import get_setting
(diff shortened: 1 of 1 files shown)


r-stein:
I also think this is useful, but would add it to jump to anywhere (https://github.com/SublimeText/LaTeXTools/pull/683) instead of to this.
I don't think it benefits from the `_jumpto_tex_file` command and in the current implementation the directive `% !TeX root = ../root.tex` would create a folder above the tex root instead of opening the root file.

ig0774:
Good points. I'll look into adding it there.

PR #764, opened 2016-07-04
Caching for BibTeX cite completions

r-stein:
This has been discussed in https://github.com/SublimeText/LaTeXTools/pull/759.

This PR adds a special caching behavior, which improves the performance of cite completions.

Two modules are added:

- `bibformat`: Utility module for formating cite completion strings
- `bibcache`: The core of this PR. A special cache for bib files.

Before caching the `bibcache` module calculates the formating strings for the cite panel and the cite autocompletion.
These results are stored in a new dict under the keys `<panel_formatted>` and `<autocomplete_formatted>`. In addition this dict keeps important keys and will be cached and given to the completion.

The completion will be check and use the preformatted entry, if a special key is present. Otherwise it will work as before. Hence it is fully backward compatible with all bibliographies.

PS:
For prefix match a match string is created and stored under the special key `<prefix_match>`.

TODO:

- [x] some small additional performance improvements
- [x] keep fields, which are needed in cite completion (for prefix match)
- [x] ST2 compatibility
- [x] doc strings for bibcache

### DIFF
--- a/latex_cite_completions.py
+++ b/latex_cite_completions.py
@@ -672,7 +672,8 @@ class LatexCiteCommand(sublime_plugin.TextCommand):
                 try:
                     return entry["<panel_formatted>"]
                 except:
-                    return [formatted_entry(s, entry) for s in cite_panel_format]
+                    return [bibformat.format_entry(s, entry) 
+                            for s in cite_panel_format]
             panel_entries = [
                 formatted_entry(completion)
                 for completion in completions
--- a/latex_cite_completions.py
+++ b/latex_cite_completions.py
@@ -31,7 +31,7 @@ if sublime.version() < '3000':
     import getTeXRoot
     from kpsewhich import kpsewhich
     from latextools_utils.is_tex_file import is_tex_file, get_tex_extensions
-    from latextools_utils import get_setting
+    from latextools_utils import bibformat, get_setting
     import latextools_plugin
 
     # reraise implementation from 6
@@ -45,7 +45,7 @@ else:
     from . import getTeXRoot
     from .kpsewhich import kpsewhich
     from .latextools_utils.is_tex_file import is_tex_file, get_tex_extensions
-    from .latextools_utils import get_setting
+    from .latextools_utils import bibformat, get_setting
     from . import latextools_plugin
 
     # reraise implementation from 6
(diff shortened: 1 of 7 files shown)


latextools_utils/bibcache.py
```diff
@@ -0,0 +1,89 @@
...
+    except:
+        raise cache.CacheMiss()
+
+    # raise a cache miss if the modification took place after the caching
+    modified_time = os.path.getmtime(bib_file)
+    if modified_time > meta_data["cache_time"]:
+        raise cache.CacheMiss()
+
+    # validate the format string are still valid
+    if any(meta_data[s] != get_setting("cite_" + s)
```
r-stein: It might be necessary to add a static version number and check `if meta["version"] == VERSION_NR and any(...`, which increases if the cached info are changed.
For example if the cite completion also needs the "author" key and changes the bibcache to keep this key, there should be a way to revoke old caches (which does not contain the key).

ig0774: That's seems like a good idea.

r-stein: Do you think it should also trigger the bibliography to be parsed or only the format the strings from the cached bibliogryphy result?

ig0774: Probably just the format strings, but I'm struggling to think of a concrete case where reparsing the bibliography is necessary. Do you have one in mind?

r-stein: No. I could image the bibliography could change and sanitize the strings, e.g. the author name from `Bj{\"o}rn` to `Björn`. Or it could somehow increase the precision, e.g. the traditional could find more authors.
However I don't see any real world example and it would not break the compatibility to the old cached entries anyway.

tlepoint:
After a few days of use, it works great for me!! 👍 It still blocks the main frame when I use new bibliography at the first compilation (building the cache), but then it is fast! 

Do you think you could improve the UI during the first build, e.g. by displaying a small window to say that it coud take some time for large bibliography files (say >6MB or so)?

ig0774:
Probably a good use for the "progress_indicator" from #732, which should be easy to integrate.

r-stein:
It think showing a "ok, cancel"-Dialog to inform the user that it might require some time (if the bib file has a certain size) and using the process indicator might be a good idea.
In general it would also be possible to move it to an other thread and not block the main thread at all like the usepackages cache does (?). However I don't know it would be worth the effort, since bib files are usually not that big.

I think this PR is ready to merge and such changes would be for the next one. Ideas and discussions are welcome anyway.

latex_cite_completions.py
```diff
@@ -783,10 +664,28 @@ def on_done(i):
...
+                ["{title} ({keyword})", "{author}"])
             # show quick
-            formatter = Formatter()
-            view.window().show_quick_panel([[formatter.vformat(s, (), completion) for s in cite_panel_format] \
-                                        for completion in completions], on_done)
+            def formatted_entry(entry):
+                try:
+                    return entry["<panel_formatted>"]
+                except:
+                    return [formatted_entry(s, entry) for s in cite_panel_format]
```
ig0774: Isn't this an infinite loop? It should probably be `bibformat.format_entry(s, entry)`

r-stein: Oh good catch.

PR #774, opened 2016-07-16
Utility function: Quickpanel for file navigation

r-stein:
This PR adds a utility function, which creates a quickpanel to navigate between file regions. It has a pretty simple interface and can be used with the analysis results.

I am looking forward for the following features:

*Full document "goto overlay":*

lt_demo_goto_section_overlay

*Show label usage in "jump to anywhere":*

lt_jumpto_label_usages

*Search for commands in document:*

lt_demo_search_commands

@msiniscalchi @ig0774  I licensed this file with the MIT license, because I think it is also useful in other projects and MIT seems to be the most permissive license.
If you also don't see a problem with the license I would merge this, such that I can create follow up PRs.

### DIFF
--- a/01_reload_submodules.py
+++ b/01_reload_submodules.py
@@ -31,6 +31,7 @@ LOAD_ORDER = [
     'latextools_utils.is_tex_file',
     'latextools_utils.sublime_utils',
     'latextools_utils.cache',
+    'latextools_utils.quickpanel',
 
     # depend on any previous
     'latextools_utils.analysis',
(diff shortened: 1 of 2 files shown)


ig0774:
I don't see a problem with a sub-component licensed under MIT. In fact we have a couple already in the `external` directory.

I'm a little hesitant about merging this into `master` only because I don't like the idea of actively-developed code in the main branch, even if it isn't used; but if its mature enough that it won't be changing too much, I don't have a problem with it being there.

Otherwise, I like the look of those features you've previewed!

r-stein:
I will add ST2 compatability before merging it. Afterwards there will be only minor changes (like adding a flag to a function).

Issue #743, opened 2016-05-24
feature request: pLaTeX 


speedcell4:

r-stein:
You should provide more information about your request. pLaTeX seems to be a LaTeX to html and pdf converter. If you want to use it as [build system](http://sublimetext.info/docs/en/reference/build_systems.html) you can easily create it for yourself. Otherwise you should write a description for your request.

speedcell4:
@r-stein sorry. here are some more information about it.

pLaTeX is a special version of LaTeX which is designed for Japanese.

> LaTeX Error: This file needs format `pLaTeX2e' [\NeedsTeXFormat{pLaTeX2e}

error will be displayed if you us jarticle .cls, but replace the `DEFAULT_COMMAND_LATEXMK` in `traditionalBuilder.py` with the following code will solve this problem.

```python
DEFAULT_COMMAND_LATEXMK = ["latexmk", "-cd",
                    "-e", "$latex = 'platex -synctex=1 -src-specials -interaction=nonstopmode'",
                    "-e", "$biber = 'biber %O --bblencoding=utf8 -u -U --output_safechars %B'",
                    "-e", "$bibtex = 'pbibtex %O %B'",
                    "-e", "$makeindex = 'makeindex %O -o %D %S'",
                    "-e", "$dvipdf = 'dvipdfmx %O -o %D %S'",
                    "-e", "$pdf_mode = '3'",
                    "-e", "$pdf_update_method = '0'",
                    "-e", "$pdf_previewer = 'open -a preview.app'",
                    "-f", "-norc", "-gg", "-pdfdvi"]
```

speedcell4:
But in that way, it will crash when turn back to use `% !TEX program = xelatex` pragma.

ig0774:
@speedcell4: A couple of things:

1. It shouldn't crash, though it won't use `xelatex`. I'm not exactly sure why you would try to use that pragma, since it's not going to do anything any way. Are you using the latest release of LaTeXTools? An earlier version (prior to v3.5.0) was deeply tied to the exact format of `DEFAULT_COMMAND_LATEXMK` and would fail if you changed it.

1. Instead of *replacing* the command in `traditionalBuilder.py`, we actually have a setting that you can use to override the command run. In your builder settings block you could use something like this (I've eliminated a few unnecessary settings):

```json
"builder_settings": {
    "command": [
        "latexmk", "-cd",
[...]

Issue #255, opened 2013-08-18
Features requested


harishkumarholla:
1. Is it possible to add a menu item and /or toolbar button for latextools in sublime text? Though adding toolbar seems too much, atleast a menu item for latextools underwhich main commands are listed will be a value addition. The menu may be made visible only when a tex file is open (like in vim-latex).
2. Is it possible to put a menu item that has drop down list of all available compilers for tex so that choosing an engine will be very easy. A similar feature is available in WinEdt (www.winedt.com) for windows. I am attaching a screen shot for the same.

First a dropdown menu in toolbar button:

![capture](https://f.cloud.github.com/assets/2251874/981936/f59e04a0-07a8-11e3-865a-d2e852da0d7e.PNG)

And a menu for TeX:

![capture1](https://f.cloud.github.com/assets/2251874/981937/17d3eb0c-07a9-11e3-83dc-c44007c581d4.PNG)

These features are really useful especially in customising the build commands. I know that I am asking toomuch but I didn't want the idea to slip by.

Thank you for the great work.

r-stein:
Now (`v3.12.0`) it is possible to see a list of different build systems via `C-shift-b`.

PR #775, opened 2016-07-16
Add variants for different build configurations

ig0774:
This adds a few build variants for some simple configurations and, more importantly, adds a way for custom `.sublime-build` files to override build settings.

Basically, this allows us to implement something not far off what was requested in #255:

screen shot

This menu can be brought up by pressing `C-shift-b`. On ST3, after build 3080, this is, of course, simply the default behaviour. However, I've back-ported something similar for earlier versions of ST3 and for ST2. The chosen build variant will persist across builds, but can be reset by choosing the **LaTeX** variant, which is simply the default configuration. (The functionality isn't *exactly* the same, but it's quite close; the main difference is that the selected build is *not* saved across windows or persisted after the application closes).

The other thing this allows is for us to support the request in #743 (see [my comment in that thread](https://github.com/SublimeText/LaTeXTools/issues/743#issuecomment-225304102)). More specifically, the following options can be provided by a `.sublime-build` file:

* `program`
* `builder`
* `command`
* `env`
* `path`

These, if provided, will override the corresponding settings, including the `%!TEX program = ...` magic comment.

We can always add more options if requested, but I thought this was a good starting point.

### DIFF
--- a/latextools_utils/sublime_utils.py
+++ b/latextools_utils/sublime_utils.py
@@ -18,6 +18,8 @@ except ImportError:
 
 __all__ = ['normalize_path', 'get_project_file_name']
 
+_ST3 = sublime.version() >= '3000'
+
 # used by get_sublime_exe()
 SUBLIME_VERSION = re.compile(r'Build (\d{4})', re.UNICODE)
 
@@ -309,7 +311,7 @@ QUOTE = re.compile(r'(?<![^\\]\\)"')
 NEWLINE = re.compile(r'\r?\n')
 
 
-def parse_json_with_comments(filename):
+def _parse_json_with_comments(filename):
     with codecs.open(filename, 'r', 'utf-8', 'ignore') as f:
         content = f.read()
 
@@ -365,3 +367,12 @@ def parse_json_with_comments(filename):
         new_content.append(content[index:])
 
     return json.loads(''.join(new_content))
+
+
+if _ST3:
+    def parse_json_with_comments(filename):
+        with codecs.open(filename, 'r', 'utf-8', 'ignore') as f:
+            content = f.read()
+        return sublime.decode_value(content)
+else:
+    parse_json_with_comments = _parse_json_with_comments
--- a/latextools_utils/sublime_utils.py
+++ b/latextools_utils/sublime_utils.py
@@ -1,5 +1,6 @@
 from __future__ import print_function
 
+import codecs
 import json
 import os
 import re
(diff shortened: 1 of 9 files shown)


latextools_utils/sublime_utils.py
```diff
@@ -200,102 +202,166 @@ def check_processes(st2_dir=None):
...
+# tokens used to clean-up JSON files
+TOKENIZER = re.compile(r'(?<![^\\]\\)"|(/\*)|(//)|(#)')
+QUOTE = re.compile(r'(?<![^\\]\\)"')
+NEWLINE = re.compile(r'\r?\n')
+
+
+def parse_json_with_comments(filename):
+    with codecs.open(filename, 'r', 'utf-8', 'ignore') as f:
+        content = f.read()
+
```
r-stein: For ST3 it might be the best to use `sublime.decode_value(content)`.

makePDF.py
```diff
@@ -75,7 +84,167 @@ def getOEMCP():
     return str(codepage)
 
 
+class LatextoolsBuildSelector(sublime_plugin.WindowCommand):
+
+	# on 3080+, the necessary features are built-in by default
+	if sublime.version() > '3079':
+		def run(self, select=False):
+			select = False if select not in (True, False) else select
+			self.window.run_command('build', {'select': select})
```
r-stein: I think it would be better to use a context in the keybinding and write an EventListener. There may be packages, which map the `ctrl+b` to something else, e.g. Sublimemacs.

ig0774: Could you explain a bit more what you have in mind? I'm not sure I follow...

latextools_utils/sublime_utils.py
ig0774: Good point.

makePDF.py
r-stein: For example write an EventListener

``` py
class SublimeVersionListener(sublime_plugin.EventListener):
    def on_query_context(self, view, key, operator, operand, match_all):
        if key != "_lt_st_build_version":
            return
        return sublime.version() <= '3079'
```
And add a context to the keybinding:
``` js
{ "keys": ["ctrl+b"],
  "context": [
    {"key": "selector", "operator": "equal", "operand": "text.tex.latex"}
    {"key": "_lt_st_build_version"}],
  "command": "latextools_build_selector"},
```

This avoids to call the command at all and we don't overwrite keybindings of other packages. I think this is a better solution than calling the other command from this one.

ig0774:
@r-stein: Thanks for the suggestions. They've been incorporated. In terms of the version context mechanism, I thought it was a good idea to make something a bit more generic. Basically, I've implemented the whole set of possible operators for the ST version number. It doesn't allow elegant support of the concept of `<=`, but its close enough.

I also added the ability to override the `script_commands` setting if using the script builder. This seems like a useful feature to cover, e.g., #828.

PR #776, opened 2016-07-17
Add ability to jump to the root from magic comment

ig0774:
Redux of #761, but using `jumpto_anywhere` and with no fancy processing. Obviously only workable if the file has been saved to disk.


r-stein:
We should add it to the "supported jump types list" in the [README](https://github.com/SublimeText/LaTeXTools/blob/v3.9.1/README.markdown#jump-to-anywhere).

PR #777, opened 2016-07-18
Hide commands if not in LaTeX files

r-stein:
This change the visibility of the commands for the command panel `C-Shift-p`. The commands will be invisible if you are not in a LaTeX file.

Currently I just made the command invisible, which are not applicable outside LaTeX files. Remaining command visible should be discussed:

- Reset user settings to default
- Create mousemap in user folder
- Build cache of LaTeX packages
- View package documentation

~~I would at least make the documentation command invisible.~~

### DIFF
--- a/delete_temp_files.py
+++ b/delete_temp_files.py
@@ -38,6 +38,10 @@ class ClearLocalLatexCacheCommand(sublime_plugin.WindowCommand):
 
 
 class DeleteTempFilesCommand(sublime_plugin.WindowCommand):
+	def is_visible(self, *args):
+		view = self.window.active_view()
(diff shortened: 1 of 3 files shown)


ig0774:
Looking at it again "View package documentation" is probably a bit of a misnomer since one might guess that it was opening the LaTeXTools documentation. That said, and speaking only for myself, I find it useful outside of LaTeX files, especially when editing bibliographies and occasionally other file types related to a TeX document.

For similar reasons, although I agree its a good idea to hide the commands that only make sense with a LaTeX document, I think its a good idea to leave the commands that are independent of a LaTeX document visible.

PR #779, opened 2016-07-19
ToC section quickpanel

r-stein:
This adds a whole document "table of content" overlay to replace the default `C-r` keybinding (opt-out). It is also accessible via `C-l, C-r`.

The entries are indented (chapter<section<subsection<...) to create a better overview and you can show and hide the labels.

TODO:
- [x] add more section commands

demo

### DIFF
--- a/Default (Linux).sublime-keymap
+++ b/Default (Linux).sublime-keymap
@@ -187,7 +187,17 @@ LaTeX Package keymap for Linux
 		"command": "insert_snippet", "args": {"name":"Packages/LaTeXTools/Text monospace.sublime-snippet"}},
 
 
-
+	// Replace the `ctrl+r` overlay with a whole document overlay as opt-out
+	// Also add this overlay to `C-l, C-r`
+	{ "keys": ["ctrl+r"],
+		"context":  [
+			{"key": "selector", "operator": "equal", "operand": "text.tex"},
+			{"key": "overwrite_goto_overlay"}],
+		"command": "latex_toc_quickpanel"},
+	{ "keys": ["ctrl+l","ctrl+r"],
+		"context":  [
+			{"key": "selector", "operator": "equal", "operand": "text.tex"}],
+		"command": "latex_toc_quickpanel"},
 
 
 		// Auto-pair ``$''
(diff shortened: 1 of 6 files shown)


ig0774:
I routinely use KOMA-script (at least when I'm not submitting anywhere)  which offers its own set of sectioning commands: `\addpart`, `\addchap`, `\addsec`, etc. These are more or less like `\part*`, `\chapter*`, etc., but also emit the corresponding `\toc` line. It would be nice to be able to match these commands as well.

It would be nice to be able to match these in the quick panel as well (currently I use a custom symbol list and modified syntax). I can see two ways of doing something like this: 1) adding some configuration (perhaps to the Advanced settings file) or 2) just hard-coding them in.

r-stein:
I think we should hard code the most used ones. I am not sure whether it is really necessary to be very flexible in this thing, because https://github.com/SublimeText/LaTeXTools/pull/781 can create a similar behavior with a few disadvantage (no indent, no label toggle).
For example one can add the keybinding to move between `addpart`, `addchap` and `addsec`:
```js
    {
        "keys": ["ctrl+r"],
        "command": "latex_search_command",
        "args": {
            "commands": [
                "addpart",
                "addchap",
                "addsec"
            ]
        },
        "context":
        [
            { "key": "selector", "operator": "equal", "operand": "text.tex.latex" }
        ]
    },
```

Nonetheless I am open to parse it flexible from the advanced settings.

ig0774:
I don't really mind hard-coding or even writing my own implementation for my own use. I imagine though that I'm not the only one who uses this bit of KOMA-script. 

It may also be worth considering adding some context to the label. We had a request for this in atom-latextools (two actually, but it's best spelt out in msiniscalchi/atom-latextools#94. I imagine we could use the second line of the quick panel to display, e.g., the rest of the context of the line or the next (non-empty) line or similar.

In any case, this is a much better replacement for the symbol list stuff. Speaking of which, we may need to merge this or remove our symbol list for sections before / when the new syntax is pulled into ST.

r-stein:
For the context: Since the document the editor automatically previews the file position we provide the same context as Auctex/Reftex for this toc movement.
We might also add this preview functionally or somethings similar to ref and bib completions as requested in that issue (this requires a change of the quickpanel, but should be possible).

Note: the preview is due to the api limitations only available for ST3

For the new Syntax: the new syntax comes with its own symbol list and it does not seem to interfer with the old one. It seems the new label and sections are not matched by the old one. I think we should keep that for ST2 compatibility and because it does not seem to harm ST3 users.

Issue #771, opened 2016-07-13
Completion for \autoref* (with asterisk)


FichteFoll:
Currently, the `\autoref*` command does not show the completion popup. You might want to add it for `\Autoref` as well.

I don't know if there are more commands like this.

ig0774:
Implemented in v3.10.0; however, I can't find where `\Autoref` is defined...

Issue #741, opened 2016-05-21
Support multiple refs in `\cref` and similar commands


excetara2:
When you type a , in `\cite` it recognizes you are using multiple citations but this same behaviour isn't included for using multiple references. If you type `\cref{ch:blah,}` and then manually try to bring up the quick panel with all the refernces it throws an error saying no reference of `ch:blah,`.

excetara2:
So I guess this was never implemented then. I thought it might be a bug in my system so wanted to check.

ig0774:
No... It's unimplemented. `\ref` and most similar commands don't actually support multiple labels.

excetara2:
Yeah that's true I think the others that I know are `autoref`, `varioref`, and `fancyref` that allow multiple citations.

ig0774:
I've implemented support for this in #782. As far as I was able to confirm, there's only a limited number of commands that actually support commas in refs, `\cref`, `\Cref`, `\cpageref` and `\Cpageref`. This is an extension `cleveref` adds to references; normally `,` is actually allowable in a label name. As far as I can tell, no other reference commands get patched for this.

ig0774:
Released in v3.10.0. Thanks for the suggestion.

excetara2:
Awesome! Give it a try soon.

PR #782, opened 2016-07-20
Unify command completions

### LINKED ISSUE #766: Cite autocomplete steals caption brace
When using the autocomplete when typing \cite{} command inside a \caption{}, cite autocompletion steals the caption's brace.

example:
One types figure, and uses ctrl+l, e to begin an environment. Then one inserts a graphic (or table, etc.) and starts writing \cap..  and use autocomplete to end up with (where the cursor is at the HERE):
```
\begin{figure}
\includegraphics{x.jpg}
\caption{HERE}
\end{figure}
```
Then one could type a caption:
`This picture was published originally by me in \cite{`
and use the autocomplete there to select the reference needed.

The result will be (with cursor at HERE)
```
\begin{figure}
\includegraphics{x.jpg}
\caption{This picture was published originally by me in \cite{thereference}HERE
\end{figure}
```
and so the cite autocompletion has eaten the closing bracket of `\caption{}`

Manually typing (so with autocompletion disabled) `\cite{thereference}` would open a new set of braces when you type `\cite{` and cause no trouble with the closing caption brace.

ig0774:
This is an overhaul of the way we handle ref / cite / input / env completions. Basically, I've unified things so that `latex_fill_all.py` handles pretty much all the completions, which each type defining a simpler plugin that simply gets the completions.

**Main features:**

* I've added a smarter bracket matching algorithm. If you type a bracket, e.g., `{`, it will automatically insert the corresponding closing bracket, e.g. `}` if the default `auto_match_enabled` setting is `True`. If you do not type the bracket (so using `C-l,f` or similar) it will automatically close the nearest bracket if it isn't already closed. This should fix #766. Because the algorithm relies on bracket counting, it could potentially perform badly on extremely large files, so it can separately be disabled by setting `smart_bracket_auto_trigger` to `False` or using the `smart_bracket_auto_trigger` toggle.

  _Smart Bracket matching:_
  citation_1
  citation_2

* In almost all cases, inserting the brackets and the completion are done in a single edit operation (as opposed to up to three), which means a single undo will undo all of them. The only exception is with autocompletions in an unclosed bracket with `smart_bracket_auto_trigger` enabled, where it is necessary to insert the closing bracket before returning the results.

* Removing the line matching code from the code that gets the list of completions allows the creation of new commands that insert a completion regardless of the current context. This is useful to, e.g., insert a bibtex key or reference in a command that LaTeXTools doesn't currently match.

* Centralising things like this allows for the creation of an `on_query_context` method to detect whether or not we are in the correct context. This enables us to stop maintaining multiple regexes for essentially the same behaviour (we had them *both* in the Python code and the keymaps). The cite regexes, in particular, are a mess and essentially unmaintainable in the keymap files. This makes it easier to address #771 and #741, which this patch also does.

* Extended `\ref` command matching. Now supports the documented reference functions of `hyperref`, `cleveref`, `fancyref` and `varioref`, including using multiple labels separated by commas in the appropriate `cleveref` commands and both arguments to `\cpagerefrange`, `\vpagerefrange` and similar commands. This is more or less what was requested in #771 and #741, but I haven't added forms I couldn't get to work (`\Autoref` and commas in non-cleveref commands).

**Additional features added:**

* `C-l,C-alt-f`: a variant of `C-l,f` that will replace the item that the caret is in with a new one. This is useful for, e.g., changing citations or references that have already been entered (it can be used with the other completions, but generally isn't necessary since they ignore the prefix).

  _Replace current option:_
  citation_3

* `C-l,alt-x,c` and `C-l,alt-x,r`: inserts a citation or ref regardless of context. This is useful to insert a citation or ref into a command that LaTeXTools currently doesn't match. Note that this keybindings do not use the bracket completion behaviour.

 _Force insert citation:_
  citation_4

**Removed features:**

* `latex_ref_cite_completions.py` - this has been completely replaced by `latex_fill_all.py`

* (undocumented) paren matching with refs and the `refp` completions. This can be added back in if there's demand.

* (undocumented) `citeX` expansion to `\cite*`. Just type `\cite*`

* the `on_insert_citation` callback for citation plugins. This could be added in, but the reason it was there in the first place seems to have been superceded by developments in Zotero Better Bibtex.


musm:
Very nice! Out of curiosity which colorscheme are you using?

r-stein:
@musm Looks like [Tomorrow](https://packagecontrol.io/packages/Tomorrow%20Color%20Schemes) or a Tomorrow based like [Aproso](https://packagecontrol.io/packages/Theme%20-%20Aprosopo)

ig0774:
It's actually "Boxy Ocean", which is included as part of the [Boxy Theme](https://packagecontrol.io/packages/Boxy%20Theme).

PR #785, opened 2016-07-23
Math Live Preview and Preview Images

r-stein:
Add the functionality to preview the currently edited math formula using phantoms and preview images using a hover popup (default) or phantoms.

Preview math:

preview_math

Preview Image:

preview_image

How does math live preview work?

1. extract the content of the current math scope
2. wrap this inside a document template
3. write this into a `*.tex` document
4. compile to document do create a pdf (using `pdflatex`)
5. convert the pdf into a png (using ImageMagick - `convert`)
6. open a html popup, which displays the generated image

Preview image features:

- images are rescaled to fit into a user defined dimension
- image types like `.pdf` are supported by creating thumbnails with ImageMagick
- provides a button to open the image and the image folder

TODO:
- [x] high dpi support
- [x] configurable latex program (not only `pdflatex`)
- [x] image tinting according to colorscheme
- [x] user configurable document template?
- [x] read several variables from the setting and listen to setting changes
- [x] check which latex math environments can be used with star (we don't want equation numbering)
- [x] clear the cache with the generated files
- [x] check, that ImageMagick is installed
- [x] ST3(b3114)/ST2 compatibility
- [x] several code improvements

- [x] test it
  - [x] on Windows
  - [x] on Linux
  - [x] on OSX

- [x] documentation/README

Future ideas:
- extract and inject used packages (?) and command definitions (?) from the edited document


msiniscalchi:
YESSS!!!!

r-stein:
Update: I will use phantoms, so it will not blink and work with all formulas to same time:

lt_preview_math_phantoms

This might also be useful for image previews (maybe with an alternative hover popup):
lt_preview_image_phantom

315234:
Doesn't seem to be able to handle unicode characters in equations:

```
  File "/Users/MyUserName/Library/Application Support/Sublime Text 3/Packages/LaTeXTools/preview_math.py", line 227, in <lambda>
    lambda: self.on_after_modified_async(view))
  File "/Users/MyUserName/Library/Application Support/Sublime Text 3/Packages/LaTeXTools/preview_math.py", line 216, in on_after_modified_async
    preview_math(view, _pos(view), live_preview=True)
  File "/Users/MyUserName/Library/Application Support/Sublime Text 3/Packages/LaTeXTools/preview_math.py", line 190, in preview_math
    _create_image(view, temp_path, base_name, latex_document)
  File "/Users/MyUserName/Library/Application Support/Sublime Text 3/Packages/LaTeXTools/preview_math.py", line 110, in _create_image
    f.write(latex_document)
UnicodeEncodeError: 'ascii' codec can't encode character '\u03d5' in position 2426: ordinal not in range(128)`
```
Note that I checked out the main repo, made a new branch and then merged your `preview_math` branch to get this installed.

315234:
Currently assumes that everyone has the same preamble. It would be good if users could specify a preamble file if they want. i.e.

```python
latex_template = """
\\documentclass[preview]{standalone}
<<preamble>>
\\begin{document}
<<content>>
\\end{document}
"""
```

and then 

```python
preamble_file = view.settings().get("tex_preamble_file", "")
if os.path.isfile(preamble_file):
    premable = open(preamble_file, "r").read()
else:
    premable = """
        \\usepackage{amsmath}
        \\usepackage{amssymb}
        \\usepackage{latexsym}
        \\usepackage{mathtools}
    """
```

r-stein:
@315234 Thanks. I moved the old branch [here](https://github.com/r-stein/LaTeXTools/tree/preview_math_popup) and pushed the new phantom version, which is based on the current master branch (`v3.10.2`). So you might try out the new phantom version.
I changed the template to:
``` tex
\documentclass[preview]{standalone}
<<packages>>
<<preamble>>
\begin{document}
<<content>>
\end{document}
```
and I am looking forward to read the packages and preamble *(command definitions...)* from the settings. It might also be useful to extract them from the current document.

315234:
I think extracting them from the current document is likely to be fragile, especially with multi tex file documents.

Currently `pdflatex` is hard coded. For seamless experience, I think before going live this should respect whatever the user has configured as the LaTeXTools `texpath` and `builder_settings["program"]`. I would love to offer a code snippet but unfortunately I couldn't figure out how to do this.

I think the default settings are still set to `"math_live_preview": true` which causes an error.

The phantoms are working really well. Using `"single"` is a really nice balance between performance and usefulness. Now if only `pdflatex` and friends worked faster...

r-stein:
It would not be to complicated to get other engines, however I am worring they don't support the `-interaction=nonstopmode` flag or require more arguments.

ig0774:
All the PDF-producing engines support `-interaction=nonstopmode` (we actually use this in the builders) and shouldn't require any more arguments that `pdflatex` does. `xetex` and `luatex` are essentially built on top of `pdftex`.

There may be some issues with requiring, e.g., a different set of packages or similar, but this could be handled by user-defined templates etc.

315234:
The following snippet will make the pngs render at 2x resolution, which looks much better on high dpi displays. It could be improved by detecting the screen resolution somewhere.

```python
    try:
        # Try to get dimensions of PNG to scale for high dpi screens
        stdout = subprocess.check_output(["identify", "-format", "%[fx:w] %[fx:h]", image_path]).decode()
        width, height = [float(s) for s in stdout.split()]
        content = '<img width={} height={} src="data:image/png;base64,{}" />'.format(width/2, height/2, img_data)
    except:
        content = '<img src="data:image/png;base64,{}" />'.format(img_data)
```

r-stein:
Wouldn't it be enough to change the density for the convert command?

315234:
Unfortunately Sublime doesn't currently draw the image at native resolution, it draws the image the same size on every platform. Rendering the image at double resolution and then scaling it down in the `` tag should work for both normal and double resolution. Sublime also doesn't currently support anything like `width=50%`, hence the call to `identify`.

As implemented, this only supports exactly double resolution (i.e. "retina" screens), not arbitrary DPI, but it should look better than normal resolution anyway.

r-stein:
I added a `_scale_quotient` variable. Together with `_density` it should support arbitrary resolutions. I will read it from the settings in the future. @315234 Can you validate, that it has the desired effect?

315234:
The new `_scale_quotient` implementation works very well. Especially nice way to get the PNG dimensions. The only thing I would add is maybe `_density = 150*_scale_quotient`, unless these are going to be user configurable (ideally).

I'm not a big fan of the way that packages and preamble are handled at the minute.

For my use case, I have over 100 lines of packages/preamble. I would rather not have to reproduce all of that in a settings file, in which many characters have to be escaped and the settings file is likely to get out of sync with the actual preamble file I am using. For me, it is much better to be able to specify the existing preamble file to use.

Package imports and preamble are not really separate things. Sometimes I have to import a package, then define some commands or settings before importing another. Having them as separate settings (and one as a list, the other as a string) seems overly complicated to me. Especially since they are just used directly after one another in the template.

r-stein:
The motivation for 2 settings is, that I assume most users use the same set of math packages for each project. However the defined math operators and commands may differ. Since each setting can be changed per project, one can just change the preamble setting and let the packages being the generic setting.
I am aware that the packages are a part of the preamble and there might be better setting names, but I think it is useful to have the possibility to logically seperate them. (Nonetheless nobody forbits to define commands inside the packages.)
It might be better do also input the preamble as a list of strings/lines.

P.S.: I want them to be settings to monitor changes and update the phantoms, after the setting has changed. This would be harder working with files.

I think for your use case it would be the easiest to make a configurable template tex file and specify the path via a setting. There you can just omit `<<packages>>` and `<<preamble>>` and write them directly into the file. This file is just read and cached and changes done to this file won't refresh the cached template.

315234:
Fair point about the settings changes. I can work around this for now by setting my preamble to `\include{whatever}`. The downside to this is that changes in the preamble file will not be reflected in the hash used for caching, and will also not be detected to update phantoms. I think having three separate settings for latex template file, packages, and preamble will be unnecessarily complex/confusing, since they will all effectively accomplish the same thing. The latex template file also wouldn't solve my problem of the hash not changing, since I would need to `include` my actual preamble file.

Personally I think the most elegant solution would be a sensible default preamble, a preamble_string setting and a preamble_file setting (with one having precedence over the other). The only issue I can see is that a preamble file would not be monitored for changes (but changes would be reflected in the hash).

Anyway, I'll stop going on about this now. Thanks for all your work on this, I am really enjoying using it.

r-stein:
I see your point and I am still thinking about the best solution. Aside: Using the preamble of the real tex document might drastically increase the tex compilation time.

preview_utils.py
```diff
@@ -0,0 +1,157 @@
...
+    global _lt_settings
+    _lt_settings = sublime.load_settings("LaTeXTools.sublime-settings")
+
+
+def convert_installed():
+    """Return whether ImageMagick/onvert is available in the PATH."""
+    if hasattr(convert_installed, "result"):
+        return convert_installed.result
+
+    convert_installed.result = shutil.which("convert") is not None
```
ig0774: This won't work on ST2 since `shutil.which()` was added in Python 3.3. The fix is to import which from `latextools_utils.system` which is identical to `shutil.which()`.

ig0774: On the other hand, will any of this work on ST2... I suppose not if it's based on phantoms.

r-stein: Yes this is ST3 only. I will add the compatibility by moving the files into subfolders als you have suggested in [the forum](https://forum.sublimetext.com/t/avoid-single-plugin-file-from-beeing-loaded-in-older-st-versions/23243)

r-stein:
@315234 I have added a setting `preview_math_latex_template_file` to specify the path to the template, which can overwrite the default template.

@ig0774 it would be nice if you could confirm, that it works on OSX

sballin:
screen shot 2016-10-21 at 10 52 18 pm

I can confirm that image and math previews work on OS X. I'm very happy with it already, but it's worth noting that math expressions appear to have problems with cropping. Also, single variables like $a$ are shifted pretty high up. Finally, I'm using vintage (vim) mode, and when I try to navigate past the end of a `\[...\]` displaymath using the `l` button, the cursor gets stuck. These may all be the result of my package ecosystem, so someone else should check.

I also used the [Fix Mac Path](https://github.com/int3h/SublimeFixMacPath) plugin so that sublime could find my imagemagick in `usr/local/bin` installed by Homebrew.

r-stein:
@sballin  Thanks for confirming OSX compatibility and your feedback.

> I'm very happy with it already, but it's worth noting that math expressions appear to have problems with cropping.

If you mean, that the long `\[ ... \]` are displayed inline and not as blocks I agree with you. My assumption was, that one would use multiple lines for long statements and only write short statements in one line. So the phantom is displayed as a block below the equation if you use multiple lines. It might be useful to 	additionally add block phantoms if the equation exceeds a certain length.

> Also, single variables like $a$ are shifted pretty high up.

I also noted this. It is because the image is such small and the phantoms are aligned with the top border (which is imo a good decision). I cannot easily change either one, so this won't change for a while.

> Finally, I'm using vintage (vim) mode, and when I try to navigate past the end of a \[...\] displaymath using the l button, the cursor gets stuck.

This is because the `l` command does not move over the line ends and phantoms are not handled as part of the editor content.

sballin:
About the cropping, I think aligning with the top is a good decision. But I'm not sure why bits of the top and bottom of the preview are getting cut off in a lot of cases. Checking for one-letter, [a-z] expressions and aligning them with the bottom of the line would be nice either way.

>This is because the `l` command does not move over the line ends and phantoms are not handled as part of the editor content.

You're right, this was silly of me, it's also affecting the `j` command but it's only happening for the expression with the box in it, in the image I posted above. Probably nothing to worry about.

r-stein:
@sballin 
> About the cropping, I think aligning with the top is a good decision. But I'm not sure why bits of the top and bottom of the preview are getting cut off in a lot of cases. Checking for one-letter, [a-z] expressions and aligning them with the bottom of the line would be nice either way.

I wasn't clear enough on this: It is not possible to control the aligning from the API. The best way I see would be to change the ImageMagick parameters to create a better image (minimal border size or something like that).

>>This is because the `l` command does not move over the line ends and phantoms are not handled as part of the editor content.
>
>You're right, this was silly of me, it's also affecting the `j` command but it's only happening for the expression with the box in it, in the image I posted above. Probably nothing to worry about.

I have created an issue for that https://github.com/SublimeTextIssues/Core/issues/1470

PR #811, opened 2016-08-10
Phantoms for build errors

r-stein:
ST adds phantom to indicate build errors in build 3118. This PR adds similar phantoms to the LaTeX build. *This is necessary, because it does not use the `exec` command.*

It looks like:

build_phantoms

It might be necessary to change the OSX and Linux regex, because they only match the message up to the first dot. I.e.:
- OSX: `^(...*?):([0-9]+): ([0-9]*)([^\\.]+)` to `^(...*?):([0-9]+): ([0-9]*)(.*)$`
- Linux: `^(...*?):([0-9]+): ([0-9]*)([^\\.]+)` to `^(...*?):([0-9]+): ([0-9]*)(.*)$`

### DIFF
--- a/makePDF.py
+++ b/makePDF.py
@@ -385,9 +385,9 @@ class CmdThread ( threading.Thread ):
 			self.caller.output("\n\n[Done!]\n")
 
 			if _HAS_PHANTOMS:
-				self.caller.errors = errors
-				self.caller.warnings = warnings
-				self.caller.badboxes = badboxes
+				self.caller.errors = locals().get("errors", [])
+				self.caller.warnings = locals().get("warnings", [])
+				self.caller.badboxes = locals().get("badboxes", [])
 
 			self.caller.finish(len(errors) == 0)
 
@@ -408,6 +408,10 @@ class make_pdfCommand(sublime_plugin.WindowCommand):
 	errs_by_file = {}
 	phantom_sets_by_buffer = {}
 	show_errors_inline = True
+	errors = []
+	warnings = []
+	badboxes = []
+
 
 	def __init__(self, *args, **kwargs):
 		sublime_plugin.WindowCommand.__init__(self, *args, **kwargs)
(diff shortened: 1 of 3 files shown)


makePDF.py
```diff
@@ -379,6 +383,12 @@ def run ( self ):
 
 			self.caller.output(content)
 			self.caller.output("\n\n[Done!]\n")
+
+			if _HAS_PHANTOMS:
```
ig0774: This block should be in the `try` block above. If the `parse_tex_log` line raises an exception, `errors`, `warnings` and `badboxes` will all be undefined raising another exception here and causing `finish()` to never be called. This also means that `errors`, `warnings` and `badboxes` probably need to be initialised in `run()` to empty lists.

ig0774:
One other things I note is that `exec.py` defines an EventListener to update phantoms on document load, which may be a good idea to emulate as well.

makePDF.py
r-stein: I will just use `self.caller.errors = locals().get("errors", [])` and additionally define them as class variables to ensure, that they are always available.

r-stein:
Agreed, I have added the event listener.

makePDF.py
```diff
@@ -672,6 +714,124 @@ def do_finish(self, can_switch_to_pdf):
...
+			if not file_regex:
+				return
+			self.errs_by_file = {}
+			try:
+				self.file_regex = re.compile(file_regex, re.MULTILINE)
+			except:
+				print("Cannot compile file regex.")
+				return
+			lt_settings = sublime.load_settings("LaTeXTools.sublime-settings")
+			level_name = lt_settings.get("show_error_phantoms")
```
ig0774: Shouldn't this just use our `get_settings()` method or is there are need to by-pass it's processing?

r-stein: I wanted to avoid to get the settings from the view, because it is applied to all views. (I could also have added the kwarg `view={}`.)

PR #837, opened 2016-09-12
Add !TEX directives to the fill all helper

r-stein:
This add the `!TEX` directives to the fill all helper, supported directives are:

- `program` show the programs `pdflatex`, `xelatex`, `lualatex` and also `pdftex`, `xetex`, `luatex`
- `root` seach current folder and parent folders for tex files
- `spellcheck` show the supported locales

Each entry is added in the auto completion menu, if you write `% !TEX`. In addition it adds a trigger on `=`.

@ig0774 This still requires minor changes in the fill all helper, because it blacklist comment scopes. It might be good to add a method `get_supported_scopes`, which returns the scopes and is `-comment` by default.

### DIFF
--- a/Default (Linux).sublime-keymap
+++ b/Default (Linux).sublime-keymap
@@ -127,7 +127,7 @@ LaTeX Package keymap for Linux
 	{	"keys": ["="],
 		"context":  [
 			{"key": "setting.disable_latex_ref_cite_auto_trigger", "operator": "not_equal", "operand": true},
-			{"key": "selector", "operator": "equal", "operand": "text.tex.latex comment.line.percentage.tex"},
+			{"key": "selector", "operator": "equal", "operand": "text.tex.latex comment.line.percentage"},
 			{"key": "selection_empty", "operator": "equal", "operand": true, "match_all": true},
 			{"key": "lt_fill_all_directive.equal_sign", "operator": "equal", "match_all": true, "operand": true}],
 		"command": "latex_fill_all", "args": {"completion_type": "directive", "insert_char": "="}},
--- a/Default (Linux).sublime-keymap
+++ b/Default (Linux).sublime-keymap
@@ -63,7 +63,7 @@ LaTeX Package keymap for Linux
 	{	"keys": ["{"],
 		"context":  [
 			{"key": "setting.disable_latex_ref_cite_auto_trigger", "operator": "not_equal", "operand": true},
-			{"key": "selector", "operator": "equal", "operand": "text.tex.latex"},
+			{"key": "selector", "operator": "equal", "operand": "text.tex.latex - comment"},
 			{"key": "selection_empty", "operator": "equal", "operand": true, "match_all": true},
 			{"key": "lt_fill_all_ref.open_curly", "operator": "equal", "match_all": true, "operand": true}],
 		"command": "latex_fill_all", "args": {"completion_type": "ref", "insert_char": "{"}},
@@ -71,7 +71,7 @@ LaTeX Package keymap for Linux
 	{	"keys": ["{"],
 		"context":  [
 			{"key": "setting.disable_latex_ref_cite_auto_trigger", "operator": "not_equal", "operand": true},
-			{"key": "selector", "operator": "equal", "operand": "text.tex.latex"},
+			{"key": "selector", "operator": "equal", "operand": "text.tex.latex - comment"},
 			{"key": "selection_empty", "operator": "equal", "operand": true, "match_all": true},
 			{"key": "lt_fill_all_cite.open_curly", "operator": "equal", "match_all": true, "operand": true}],
 		"command": "latex_fill_all", "args": {"completion_type": "cite", "insert_char": "{"}},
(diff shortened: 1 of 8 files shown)


ig0774:
@r-stein: That sounds like reasonable feature to add.

r-stein:
@ig0774 it would be nice, if you could review the changes on the fill all completions.

Default (Linux).sublime-keymap
```diff
@@ -124,6 +124,14 @@ LaTeX Package keymap for Linux
 			{"key": "lt_fill_all_env.open_curly", "operator": "equal", "operand": true, "match_all": true}],
 		"command": "latex_fill_all", "args": {"completion_type": "env", "insert_char": "{"}},
 
+	{	"keys": ["="],
+		"context":  [
+			{"key": "setting.disable_latex_ref_cite_auto_trigger", "operator": "not_equal", "operand": true},
+			{"key": "selector", "operator": "equal", "operand": "text.tex.latex comment.line.percentage.tex"},
```
ig0774: On ST2, the scope for these directives is `comment.line.percentage.directive.texshop.tex`, which means that the autocompletions do not work. It should probably be changed to either `comment.line.percentage` or just `comment`.

latex_directive_completions.py
```diff
@@ -0,0 +1,222 @@
...
+    ]
+    engines = [e for e in engines if e.startswith(value)]
+    if ac:
+        comp = [(e + "\ttex-program", e) for e in engines]
+    else:
+        comp = engines, engines
+    return comp
+
+
+_LINE_RE = re.compile(
```
ig0774: This definition doesn't quite match the definition it `get_tex_root`. In particular, `get_tex_root` supports multiple percent signs (I've come across documents with `%%% !TEX root = ...`, etc.) and it doesn't support space before the percent sign. I'd actually be happy for us to add support for spaces before the `%`; it just should be handled consistently.

latex_directive_completions.py
```diff
@@ -0,0 +1,222 @@
...
+        if prefix:
+            line_str = line_str[:-len(prefix)]
+
+        # circumvent completion if it cannot be possible
+        if "!" not in line_str:
+            return
+
+        comp = None
+
+        tex_directives = ["root", "spellcheck", "program"]
```
ig0774: It would be good to also include `output_directory`, `aux_directory`, `jobname`, and `options` here. Even if we're not supporting Fill All completions here, it would be useful (and trivial) to list them here.

latex_directive_completions.py
```diff
@@ -0,0 +1,222 @@
...
+        return get_setting("tex_directive_auto_trigger", True)
+
+
+class LatexDirectiveCompletion(sublime_plugin.EventListener):
+    def on_query_completions(self, view, prefix, locations):
+        if len(locations) > 1:
+            return
+        point = locations[0]
+        if not view.score_selector(
+                point, "text.tex.latex comment.line.percentage.tex"):
```
ig0774: Unlike the other uses of the `comment.line.percentage.tex` this actually does the right thing on ST2.

ig0774:
@r-stein: I think you've implemented the `get_supported_scopes` thing in as clean a way as is possible. However, the main point in filtering out comment scopes was just to ensure that we didn't auto-trigger when someone was writing a comment. I don't think it matters so much if someone uses `C-l,C-f` or similar to actually request a completion. In light of that, I'm wondering if it just makes sense to change the scope in the keymaps to `text.tex.latex - comment` or something. This also has the added advantage of not triggering the `on_query_completion` stuff at all when in a comment.

latex_directive_completions.py
```diff
@@ -0,0 +1,222 @@
...
+        if "!" not in line_str:
+            return
+
+        comp = None
+
+        tex_directives = ["root", "spellcheck", "program"]
+        if re.match("\s*%\s*!$", line_str):
+            row, _ = view.rowcol(point)
+            # do this completion only in the first 10 lines
+            if row < 10:
```
ig0774: Just for historical purposes, I'd make this the first 20 lines. (Not an arbitrary number; TeXShop originally only supported these comments if they occurred in the first 20 lines of the file).

latex_directive_completions.py
```diff
@@ -0,0 +1,222 @@
...
+            line_str = line_str[:-len(prefix)]
+
+        # circumvent completion if it cannot be possible
+        if "!" not in line_str:
+            return
+
+        comp = None
+
+        tex_directives = ["root", "spellcheck", "program"]
+        if re.match("\s*%\s*!$", line_str):
```
ig0774: As ludicrous as it might seem, its probably a reasonable idea to pre-compile the two regexes in this function. Also, [see my previous comment](https://github.com/SublimeText/LaTeXTools/pull/837/files#r78582993) on supporting spaces, etc.

latex_directive_completions.py
r-stein: Okay I will take a look at tex_directive and change the regex to be consistent.

r-stein:
We need the `get_supported_scopes` function at least for auto completion. However I agree, that it might be better to remove if from the triggered command (and as a scope to the keymap to optimize).

PR #865, opened 2016-10-05
jump out from brackets

YsHaNgM:
use "enter" to jump out from brackets, <>, $$


r-stein:
Thank you for the PR, I don't think that this feature is a benefit for the majority of users.

PR #878, opened 2016-10-23
encode tex_dir for Popen, to handle non ANSI characters in it

anamewing:
In ST2, Popen can not handle non ANSI characters in cwd parameter with Unicode type.
This patch fixes a build error if the path of tex file has non ANSI characters, eg. CJK characters, on windows.

I have not tested this in ST3.

### DIFF
--- a/makePDF.py
+++ b/makePDF.py
@@ -139,12 +139,13 @@ class CmdThread ( threading.Thread ):
 					# Now create a Popen object
 					try:
 						if self.caller.plat == "windows":
+							encodedTexDir = self.caller.tex_dir.encode(sys.getfilesystemencoding())
 							proc = subprocess.Popen(
 								cmd,
 								startupinfo=startupinfo,
 								stderr=subprocess.STDOUT,
 								stdout=subprocess.PIPE,
-								cwd=self.caller.tex_dir
+								cwd=encodedTexDir
 							)
 						elif self.caller.plat == "osx":
 							proc = subprocess.Popen(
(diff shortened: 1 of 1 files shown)


r-stein:
On which OS (Windows, Linux, OSX) with which latex distro (miktex, texlive, mactex) have you tested it?

Aside: please use `under_score` instead of `camelCase` variable names.

anamewing:
I only tested it on Windows, with texlive. 
The previous codes worked well on my laptop using OSX with mactex.

This issue has nothing to do with latex. It looks like a bug in ```subprocess.Popen``` on windows.

r-stein:
Okay I did not look in the code, that it is Windows only. The error also seem to only occur with ST2.
So I would suggest to change the lines to something like:
``` py
encoded_tex_dir = self.caller.tex_dir
if not _ST3:
    encoded_tex_dir = encoded_tex_dir.encode(sys.getfilesystemencoding())
```

anamewing:
Fixed.

There seems to be a new feature in github, that repository maintainer can directly commit to the pull request branch.

r-stein:
Nice to know, I can even edit the file online =)

This will be added to the next release.

anamewing:
Thanks
:)

ig0774:
Thanks for this. It looks really useful... In fact, we probably need to incorporate this change into several other places (the basic builder, the script builder, texcount, etc.). As such, it may be best to apply this change in #673, which is already slated for the next major release.

BTW, @anamewing: how would you like to be credited?

anamewing:
This is really a simple workaround, so don't be bothered for the credit.
When the #673 finishes, this simple fix maybe useless, and we can close it.

r-stein:
I will close this in favor of https://github.com/SublimeText/LaTeXTools/pull/673, since it is already commited to that PR.

PR #890, opened 2016-10-31
Preview: Fix overly close math preview cropping

sballin:
Give proper padding to math preview's document template and fix too-high baselines of smaller characters. Also remove redundant `-trim` operation in `convert` command.

### DIFF
--- a/st_preview/preview_math.py
+++ b/st_preview/preview_math.py
@@ -128,6 +128,8 @@ def _create_image(latex_program, latex_document, base_name, color,
             '-density', '{density}x{density}'.format(density=density),
             # change the color form black to the user-defined
             '-fuzz', '99%', '-fill', color, '-opaque', 'black',
+            # trim the content to the real size     
+            '-trim',
             pdf_path, image_path
         ], shell=sublime.platform() == 'windows')
 
--- a/st_preview/preview_math.py
+++ b/st_preview/preview_math.py
@@ -128,8 +128,12 @@ def _create_image(latex_program, latex_document, base_name, color,
             '-density', '{density}x{density}'.format(density=density),
             # change the color form black to the user-defined
             '-fuzz', '99%', '-fill', color, '-opaque', 'black',
-            # trim the content to the real size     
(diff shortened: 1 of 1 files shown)


r-stein:
Thanks for the PR. The problem with removing `-trim` is that e.g. math environments, e.g. `\begin{align}...\end{align}` appears centered instead of left aligned. This seem to persists with your changes.

Would you be so kind and explain the usage of of `$\\mathstrut$`?

sballin:
Ah my bad. `$\\mathstrut$` ensures that the height of the math mode statement is at least the height of `$($`. So little variables like `$a$` won't be a higher baseline anymore. You or I would need to replace the trimming on all sides with trimming in the horizontal direction only, I can look into this soon.

r-stein:
Trimming horizontal is a nice idea. If you replace the trim command by 
``` py
        execute_command([
            'convert',
            # set the image size/density
            '-density', '{density}x{density}'.format(density=density),
            # change the color form black to the user-defined
            '-fuzz', '99%', '-fill', color, '-opaque', 'black',
            # trim the content to the real size
            '-background', 'black',
            # trim the left side
            '-gravity', 'East', '-splice', '1x0', '-trim', '+repage',
            # trim the right side
            '-gravity', 'West', '-splice', '1x0', '-trim', '+repage',
            pdf_path, image_path
        ], shell=sublime.platform() == 'windows')
```

It only trims horizontal. However it looks like some `\begin{align}...\end{align}` adds an empty line above.

sballin:
Nice job figuring out the horizontal cropping! I moved the `\mathstrut` inside the content, so now it won't be adding a blank line before non-inline math. One last thing is that `convert ... -trim` seems to be lopping off an extra pixel or two on the right and left sides, but I couldn't figure out how to fix that and it's looking very good anyway.

r-stein:
I close this, because it is not compatible anymore and some changes (like border) are already on the master. In addition the problem is mostly solved now.
If you want to you can rebase and reopen the PR.

PR #906, opened 2016-11-02
Cache the result of `_get_convert_command`

r-stein:
This adds a caching for the result of the `_get_convert_command`.

- it does not change, except from the edge case of installing ImageMagick while running ST (just restart ST)
- I assumed `convert_installed` to be constant time (it takes 0.05 - 0.2 s on my machine w/o caching)

### DIFF
--- a/st_preview/preview_utils.py
+++ b/st_preview/preview_utils.py
@@ -15,11 +15,15 @@ _lt_settings = {}
 
 
 def _get_convert_command():
+    if hasattr(_get_convert_command, "result"):
+        return _get_convert_command.result
+
     texpath = get_texpath() or os.environ['PATH']
-    return (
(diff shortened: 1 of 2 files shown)


ig0774:
Doesn't this still cache the value if it winds up being `None` though? I agree that caching here is a good idea.

r-stein:
Yes, it also caches `None`. I think in the edge case of installing ImageMagick while running ST one can just restart ST. You also need to restart ST if you change the path, so this should be done anyway.
The only problem I see is that the system check might succeed while `None` is still cached. We may reset the cache if the tex_path setting changes or something like that.

ig0774:
Actually, on reflection, you're right. This is good.

r-stein:
I have added the system check to show "restart required" for that case.

Issue #910, opened 2016-11-03
Latex-cwl seems not compatible with the updated latextools


allenryb:
Hi, everyone,

It seems the items in latex-cwl can't appear in the popup, for example, the "\displaystyle" and "\frac{num}{dem}".
There are some mistakes in the popup. When I type "\infty" and click the item in the popup, the result is "\int".

Is there some method to fix this?
Thank you very much!

r-stein:
Have you installed the [LaTeX-cwl](https://packagecontrol.io/packages/LaTeX-cwl) package?

> There are some mistakes in the popup. When I type "\infty" and click the item in the popup, the result is "\int".

The idea behind this is a toggle between the different completions [LaTeX math.sublime-completions](https://github.com/SublimeText/LaTeXTools/blob/v3.12.4/LaTeX%20math.sublime-completions#L66). Although I agree, that this should improve.

allenryb:
Yes, I have installed the Latex-cwl package. Actually, after the latextools updated, I found the latex-cwl package doesn't work. I also tried to reinstall the latextools and latex-cwl. It seems no work. 

> Have you installed the LaTeX-cwl package?

I'm a new guy here. Thank you for your reply and the wonderful work! @r-stein

r-stein:
Do I understand correct the cwl completions did work with LaTeXTools, but stopped after the latest update?
Have you restartet Sublime Text then?

allenryb:
Yes, the cwl completions works well with latextools v.3.11.2. But it doesn't work in latextools v.3.12, even if I restarted the Sublime Text 3.

Now I reinstall the latex v.3.11.2 and it works.

ig0774:
@allenryb: Thanks for the bug report. This will be fixed in the next release.

r-stein:
@ig0774 it seems `command_to_snippet` returns a tuple and we wrap in [own_completions](https://github.com/SublimeText/LaTeXTools/blob/v3.12.4/latex_own_command_completions.py#L83) again.

ig0774:
@r-stein: Actually, that doesn't cause this issue. The problem was that we were calling `startswith` on a value that's always a tuple... The other issue (which caused a `Type Error: String required` message to be written to the console) is addressed in #912.

r-stein:
[...]

PR #911, opened 2016-11-03
Fix for #910

ig0774:
Simple bug fix for a bug accidentally introduced in v3.12.3.


r-stein:
I close this in favor of #913, because this is still buggy with self defined commands, that use arguments.

PR #913, opened 2016-11-03
Fixed a bug, which disabled cwl completion

### LINKED ISSUE #910: Latex-cwl seems not compatible with the updated latextools

r-stein:
Fix #910


r-stein:
Closed in favor of #912.

PR #924, opened 2016-11-05
Add some checks for packages required by preview

ig0774:
Checks for the existence of packages explicitly used by the equation preview. Note that this doesn't check the availability of the dependencies of the used packages. This is intended to aid with troubleshooting the preview function.

I've also added extensive documentation on installing ImageMagick and Ghostscript for all three platforms. I've added it to this PR as it makes reference to the functionality added here.


README.markdown
```diff
@@ -1065,3 +1098,27 @@ The second exception has to do with file and path names. In order to accommodate
 
 Finally, I have done my best to accommodate non-ASCII characters in logs. I cannot promise that everything works, but I'd like to know if you see issues with this.
  
+### Equation preview shows "ERROR: Failed to run 'pdflatex' to create pdf to preview."
```
r-stein: I am preparing a PR to improve and open the error report so this section might not be necessary.

ig0774: That sounds good to me!

PR #925, opened 2016-11-05
Preview: Improve error reports

r-stein:
Help people with debugging math-live preview by access to more information

- Add more information to the error report file, this is
  + the original tex file
  + the parsed log file (errors + warnings)
  + the raw log file
- Add a button to run the system check or show the error report file to the math phantoms (if the build fails)

### DIFF
--- a/st_preview/preview_math.py
+++ b/st_preview/preview_math.py
@@ -267,6 +267,12 @@ def _generate_error_html(view, image_path, style_kwargs):
         content += f.readline()
 
     html_content = html.escape(content, quote=False)
+    html_content += (
+        '<br>'
+        '<a href="check_system">(Check System)</a> '
+        '<a href="report-{err_file}">(Show Report)</a>'
+        .format(**locals())
+    )
 
     html_content = _wrap_html(html_content, **style_kwargs)
     return html_content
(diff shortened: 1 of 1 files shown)


st_preview/preview_math.py
```diff
@@ -143,15 +144,51 @@ def _create_image(latex_program, latex_document, base_name, color,
...
+                errors = warnings = []
+            if errors:
+                err_log.append("Logged errors:")
+                err_log.extend(errors)
+            if warnings:
+                err_log.append("Logged warnings:")
+                err_log.extend(warnings)
+            err_log.append("")
+
+            log_content = log_data.decode("utf8")
```
ig0774: Might be good to make this `log_data.decode('utf-8', 'ignore')`. Technically the log files are written using LaTeX's output-encoding. For LuaLaTeX and XeLaTeX this is always UTF-8, but for PdfLaTeX, its usually T1 which is *mostly* similar to `utf-8` (at least for the Latin alphabet) but sometimes differs.

st_preview/preview_math.py
```diff
@@ -143,15 +144,51 @@ def _create_image(latex_program, latex_document, base_name, color,
...
+        err_log.append("")
+        err_log.append("")
+        err_log.append("LaTeX document:")
+        err_log.append("-----BEGIN DOCUMENT-----")
+        err_log.append(latex_document)
+        err_log.append("-----END DOCUMENT-----")
+
+        err_log.append("")
+
+        log_file = os.path.join(temp_path, base_name + ".log")
```
ig0774: Might it be better to show the (parsed) errors before the document? Assuming the parsing code mostly does its job, this would mean that the file starts with what is probably the most relevant information.

r-stein: I agree, that would be better.

PR #945, opened 2016-11-18
Added "newacronym" in glo_filter_commands

xrayn:
According to the Documentation of the glossaries latex package, \gls{}
command is used for both, the glossary entries and the acronyms.
Although the acronyms are handled by the \acr[full|short|long] command
in the plugin, these do not support that the first usage of the acronym
is written in full form and afterwards in short. Thus, \acr* commands
needs manual handling in latex, which is usually not desired.

### DIFF
--- a/jumpto_anywhere.py
+++ b/jumpto_anywhere.py
@@ -178,7 +178,7 @@ def _jumpto_glo(view, com_reg, pos, acr=False):
     ana = analysis.analyze_document(tex_root)
     if not acr:
         commands = ana.filter_commands(
-            ["newglossaryentry", "longnewglossaryentry"])
+            ["newglossaryentry", "longnewglossaryentry", "newacronym"])
     else:
         commands = ana.filter_commands("newacronym")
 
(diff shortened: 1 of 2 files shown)


ig0774:
My only concern here is whether we should make the formatting for acronyms displayed through the `\gls` commands match that displayed via the `\acr{}` commands. Otherwise, this looks like a good idea.

r-stein:
The acronyms are 2 lines per item and glossaries are 1 line per item and I am not sure whether we should extend this because of acronyms.

r-stein:
@xrayn Would you mind to add it also to the jump-to-anywhere command [here](https://github.com/SublimeText/LaTeXTools/blob/v3.12.8/jumpto_anywhere.py#L181)?

xrayn:
@r-stein: OK, I added it. Actually, I wasn't aware of this functionality. So it took some time to figure out what key+mouse combination triggered this _jumpto_glo in the first place. :)

PR #947, opened 2016-11-19
Preview: Use ghostscript directly for equation previews

ig0774:
The intention here is to reduce the amount of configuration necessary to get equation previews working, since Ghostscript is usually installed alongside TeX.

This isn't as much help if you're also using the image preview on Windows which also requires an install of Ghostscript, but I've updated the README to include information on how to get ImageMagick to use the version of Ghostscript installed with the TeX distribution.

Finally, because this uses Ghostscript rather than ImageMagick to do the trimming, the results may be slightly different from the current version. However, it seems to do the right thing with regard to inline single-character uses and multiline equations. Admittedly, though, I haven't tried anything too complex.

Some examples:

* equation with fraction

* single letter

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -78,9 +78,15 @@ in the Command field, and `"%file":%line` in the Arguments field. (This is corre
 
 #### Setup ImageMagick and Ghostscript
 
-If you are using Sublime Text 3 version 3118 or later and want to use the image and equation previewing features, you will need to ensure that both ImageMagick 6 or higher and Ghostscript 9 or higher are installed and available on your machine. If you installed the full MacTeX distribution, Ghostscript is already included. If you installed the BasicTeX distribution, you will need to install Ghostscript yourself.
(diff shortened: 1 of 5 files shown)


r-stein:
This works nice.

The image preview could also work with ghostscript only for pdf and eps files, which are the most common file types that need to be converted. However we should keep the convert command for other file types like svg.

ig0774:
Yes, that's very true, but it might be a bit messier to handle because of the reliance on `convert_installed()`.

r-stein:
It adds a few more cases based on the extension, but should not be to complicated and we can add it in a later PR.

ig0774:
@r-stein: Turns out to be easier than I thought to use ghostscript for image previews. I was worried about trimming, etc., but since we don't do that anyway, its pretty straight-forward. Anyways, since it requires basically all the same code, I've added that to this PR.

PR #952, opened 2016-11-20
Move documentation to readthedocs

ig0774:
As many people have commented the README is a bit of an undecipherable wall of text. This aims to fix this situation by drastically simplifying the README (the proposed version of the README can be found [here](https://github.com/ig0774/LaTeXTools/blob/read_the_docs/README.markdown) and moving the documentation into a more browsable and searchable format using the fantastic [readthedocs](https://readthedocs.org) platform. The draft of this documentation (updated to v3.12.8) can be found [here](https://latextools.readthedocs.io).

I'd appreciate any feedback on how the documentation could be made clearer, better broken up, etc.


r-stein:
I think this is a huge improvement for the readability, but some things should be restructured in following PRs. E.g. "Alternative Viewers" is a top item and "Viewers" is subitem of "Features", which is kinda unititive. We should move some items of Features to the top and move "Alternative Viewers" as a subitem of "Viewers".

r-stein:
I think we should relase this with the next minor version (`3.13.0`).

We could also add a section "Additional Packages" with package recommendation of packages, which are compatible with LaTeXTools and

- are used by LaTeXTools (LaTeX-cwl, Dictionaries)
- are specific to work with LaTeX (better quotes, additional snippets, ...)
- also support the LaTeX structure with their features (BracketHighlighter, ExpandRegion, ...)
- or are just good for writing text or handling multiple-seletions

Especially users, who use ST only for LaTeX, may not be familiar with the packages. A list of packages, which are useful for writing LaTeX documents, may be helpful.

ig0774:
Another small stumbling block here is that I can't actually setup a [webhook](https://docs.readthedocs.io/en/latest/webhooks.html#github) for this repo, which would allow the documentation to be automatically rebuilt. I think this is something @msiniscalchi or someone from the SublimeText organisation would have to do.

PR #955, opened 2016-11-23
Several improvements to loading plugins

ig0774:
1. As suggested by @r-stein, plugins are loaded on a non-main thread. This means that plugin loading should interfere less with start-up time.
1. Removed the automatic support for files with custom extensions. This should also marginally reduce the plugin load time.
1. Fixed the reloading code to work properly on ST2 (FillAll plugins would routinely be left out).
1. Made `get_setting()` thread-safe on ST2. This means we should be able to run more things on separate threads where it makes sense to do so.

### DIFF
--- a/01_reload_submodules.py
+++ b/01_reload_submodules.py
@@ -12,6 +12,8 @@ import traceback
 if sys.version_info >= (3,):
     from imp import reload
 
+_ST3 = sublime.version() >= '3000'
+
 
 def _load_module_exports(module):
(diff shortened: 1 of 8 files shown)


01_reload_submodules.py
```diff
@@ -73,6 +72,9 @@ def _load_module_exports(module):
     'latex_fill_all'
 ]
 
+if sublime.version() >= '3000':
```
PeterPablo: `if _ST:`?

PR #956, opened 2016-11-23
Use analysis for Ref and Cite commands

ig0774:
TL;DR With this PR ref and cite completions use the cached analysis of the document which is asynchronously updated on save.

Essentially, I wanted to use the analysis to provide reference and citation completions, which is pretty straight-forward thanks to @r-stein's wonderful work. The problem is that in the current state, the analysis cache is only updated around once every 30 minutes, which is fine for most uses,  but becomes annoying when adding a label to a document as it may not be available immediately to be used from referencing commands.

There are a few simpler solutions to this problem:
1. Live with a 30-minute span between adding a label and LaTeXTools "seeing" it. This seems like it would get complaints from a usability perspective.
1. Ratchet down the default value for the cache lifespan. This seems like it would negate the value of caching increasingly, especially if it got to small enough values that no one complained any more.
1. Use a non-cached analysis to get the label completions. This would probably perform worse than the current implementation, especially on very large documents.
1. Stick with what we currently have. I'm very open to this, but I think this PR could ultimately make the analysis more useful.

This does three main things:
1. The cache is re-written to provide an in-memory cache, which defers writing to disk to a background thread. These in-memory caches share state across multiple instances that are the "same", so it effectively appears as though there is a single global cache, a single cache for each document (i.e., series of files sharing a `tex_root`), etc.
1. It updates the analysis so that it can be used in a background thread on ST2.
1. It adds a listener which supports running the analysis in the background when a document is loaded and / or when it is saved, so that the in-memory copy of the analysis is kept in sync with the document. These listeners can be disabled via preferences.

This means, in most cases, LaTeXTools will have a fresh copy of the analysis to work with without blocking the GUI and most cache reads and writes don't immediately involve disk access.

### DIFF
--- a/latextools_utils/cache.py
+++ b/latextools_utils/cache.py
@@ -18,11 +18,13 @@ import sublime
 if sublime.version() < '3000':
     _ST3 = False
     from latextools_utils import get_setting, utils
+    from external.frozendict import frozendict
     from latextools_utils.six import unicode, long, strbase
     from latextools_utils.system import make_dirs
 else:
     _ST3 = True
     from . import get_setting, utils
+    from ..external.frozendict import frozendict
     from .six import unicode, long, strbase
     from .system import make_dirs
 
@@ -137,6 +139,13 @@ class Cache(object):
         if result is _invalid_object:
             raise CacheMiss('{0} is invalid'.format(key))
 
+        # return a copy of any objects
+        try:
+            if hasattr(result, '__dict__') or hasattr(result, '__slots__'):
+                result = copy.copy(result)
+        except:
+            pass
+
         return result
 
     def has(self, key):
@@ -171,6 +180,13 @@ class Cache(object):
             pickle.dumps(obj, protocol=-1)
         except pickle.PicklingError:
             raise ValueError('obj must be picklable')
+
+        if isinstance(obj, list):
+            obj = tuple(obj)
+        elif isinstance(obj, dict):
+            obj = frozendict(obj)
+        elif isinstance(obj, set):
+            obj = frozenset(obj)
 
         with self._write_lock:
             self._objects[key] = obj
--- a/latextools_utils/cache.py
+++ b/latextools_utils/cache.py
@@ -68,6 +68,112 @@ def hash_digest(text):
     return hash_result.hexdigest()
 
 
+def cache(tex_root, key, func):
+    '''
+    alias for cache() on the LocalCache instance corresponding to the tex_root:
+
+    convenience method to attempt to get the value from the cache and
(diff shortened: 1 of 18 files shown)


r-stein:
We should also create a copy of the object via `pickle.loads(pickle.dumps(self._objects[key], -1))` in the get method, because currently it is ensured, that you get your own object instance and you can do whatever you want.

r-stein:
I would suggest to store the pickled object in `self._objects[key]` (we need to pickle it anyway to write it into the file) and unpickle it in each get. We may add a flag `ALLOW_CACHE_SAME_INSTANCE` for saving and avoid pickling then to have a better performance.

ig0774:
@r-stein: The approach I've taken is somewhat different, though I can easily back it out and go with pickled objects. The problem with pickling and de-pickling on each get is that its fairly substantial overhead and storing pre-pickled strings (which cuts the overhead roughly in half) gets very complex (i.e. they have to be de-pickled on write and perhaps separately re-pickled on load or something along those line).

My alternative was to freeze the analysis once it was finished, the values returned were essentially immutable (the only thing that's still mutable is the `sublime.Region`s), which seems to also eliminate the issue. I did something similar with the bib_cache, though this was a bit more straight-forward to do.

r-stein:
I am fine with freezing the analysis und the bibcache, but we should also ensure that other things we add to the cache are immutable (e.g. lists or dicts). I have no case where this would really result in a problem, but future debugging could be harder.

latextools_utils/cache.py
```diff
@@ -55,313 +68,577 @@ def hash_digest(text):
...
-def invalidate_local_cache(cache_path):
-    """
-    Invalidates the local cache by removing the cache folders
-    """
-    if os.path.exists(cache_path):
-        print(u"Invalidate local cache '{0}'.".format(repr(cache_path)))
-        shutil.rmtree(cache_path)
-
-
-def cache(tex_root, name, generate):
```
r-stein: I would like to keep `cache`, `read`, and `write` to easy access the local cache.

ig0774:
@r-stein: I agree we should try to keep things as immutable as possible. So, as far as possible the changes also try to use frozen versions (so dicts are converted into an immutable dict and lists to tuples). In addition, objects are copied (using shallow copies) on retrieval, which gives us an opportunity to provide custom handling of the copy where needed (so, for example, Analysis, which freezes itself, now just returns the same instance)

I've also re-added `cache` and `cache_global` as well as corresponding `get` and `set` functions. I've renamed them for some consistency in terminology, since `get` and `set` are the names used in the classes and `_read` and `_write` are internal methods referring to actual disk access.

PR #958, opened 2016-11-23
Fixed bug on opening external pdf viewer

trukanduk:
In some cases _command becomes a None. Also check a command has no sense: command[0] already exists, so i think it was a typo.


ig0774:
Indeed it was. Thanks!

Issue #861, opened 2016-10-01
tex option --shell-escape not working


wangzhengbeijian:
I use package minted to compile my tex, which requires shell-escape option. So I add that directive at the top of file, but it doesn't work.

``` latex
%!TEX program = xelatex
%!TEX options = --shell-escape
\documentclass{article}
\usepackage{minted}
\begin{document}
\end{document}
```

And finally I find the solution. At line 128 at file 'tradiotionalBuilder.py', I change \" to ', and it works. 

``` python
for option in self.options:
    if texify:
        cmd.append(u"--tex-option=\"" + option + "\"")
    else:
        cmd.append(u"-latexoption='" + option + "'")
```

I'm not familiar with python, and don't know why. Hope you can do some change to tradiotionalBuilder.py to solve the problem, may not according to my solution.

ig0774:
Thanks for pointing this out. I thought this was working already, but clearly I missed something. What OS are you on?

wangzhengbeijian:
Windows 7, Sublime Text 3126

ig0774:
This has been fixed in the most recent release.

Sharparam:
I also noticed the `%!TeX options = ...` not working properly recently. The fix that was implemented didnt' fix it for me though. If I remove the quoting entirely, it works. So making it look like this:

```
for option in self.options:
  if texify:
    cmd.append(u"--tex-option=\"" + option + "\"")
  else:
    cmd.append(u"-latexoption=" + option + "")
```

With the quotes in, latexmk parses it as a file and complains about not finding it: "I can't find file '-shell-escape'".

I'm not sure if the quotes are required for other types of options, but for `-shell-escape`, quoting it breaks it.

Gitsack:
This is still not working.

```Latexmk: applying rule 'pdflatex'...
This is pdfTeX, Version 3.14159265-2.6-1.40.17 (TeX Live 2016/W32TeX) (preloaded format=pdflatex)
 restricted \write18 enabled.
entering extended mode
! I can't find file `'--shell-escape''.
<*> '--shell-escape' 
```

How can I fix this?

Sharparam:
@Gitsack See my comment above yours, you have to make it not quote the options.

I haven't made a PR to change this as I'm unsure if the quoting might be necessary for other options.

Gitsack:
@Sharparam thanks for your help, it works now.

[...]

PR #983, opened 2016-12-11
Remove quotes from string passed to latexmk

ig0774:
See both issues raise in #861


raimund-schluessler:
This solves the issue in #861 for me.

PR #984, opened 2016-12-11
Added keyboard shortcut for small caps text (Ctrl+L, Ctrl+S)

morgil:
Well, I often need to write small caps text, so I wrote a shortcut for this and want give it to the community. I propose Ctrl+L, Ctrl+S.


r-stein:
Everything looks fine, but I am hesitating a little bit with adding additional C-l keybindings. Some day we may want to use that for something else.

igreil:
Perhaps not the most pressing issue, but can we review this issue? Unless I am missing something, this is still not an option, and writing \textsc{ ... } by hand every time becomes old, fast.

deathaxe:
Well, these bindings now conflict with recently merged #1424.

PR #985, opened 2016-12-12
Preview: Provide clearer equation previews

ig0774:
This works by adding the `dTextAlphaBits` and `dGraphicAlphaBits` settings recommended in the [Ghostscript documentation](http://www.ghostscript.com/doc/9.20/Devices.htm#File_formats). In addition, it adds a new setting `"preview_math_hires"` which, if set to `true` (default), the image is rendered at 8x the requested density and then scaled down (by Ghostscript) to the requested size. This seems to result in a clearer image. Finally, I've changed the default settings for the preview density and scale factor as image scaling in ST seems to add a small amount of distortion.

Over all, this seems to make previews crisper. For whatever reason, the current distortion seems to be more pronounced on Windows and Linux than on macOS, at least for me, and hence this has a bigger impact on Windows or Linux users.

A quick example, using the generated PNGs.

Current state:

PNG from Linux

After change:

PNG from Linux

New Rendering in action:

screenshot showing new rendering

### DIFF
--- a/LaTeXTools.sublime-settings
+++ b/LaTeXTools.sublime-settings
@@ -142,10 +142,13 @@
 	"preview_math_no_star_envs": [],
 
 	// The density of the preview image. The higher the density the bigger the phantom.
-	"preview_math_density": 300,
+	"preview_math_density": 150,
 	// If the image is not sharp enough increase this scale to get a better resolution.
 	// However also change the density by the same factor to keep the size.
-	"preview_math_scale_quotient": 2,
+	"preview_math_scale_quotient": 1,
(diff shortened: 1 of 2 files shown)


r-stein:
Please in crease the [_version](https://github.com/ig0774/LaTeXTools/blob/073717dea432891ba62bfe053adcbec35614da06/st_preview/preview_math.py#L28) number to invalidate old images.

ig0774:
@r-stein: I added the value of `hires` into the hash so it will update when the value changes. This will also invalidate any existing images.

r-stein:
Okay that is also fine.

PR #990, opened 2016-12-27
Support for the import package

r-stein:
This PR adds support for the commonly used import package.

TODO list
- [x] include it into the analysis
- [x] fix input/include/includegraphics completions
- [x] add `\import` command completion
- [x] fix input/include/includegraphics in jumpto-anywhere
- [x] support it in the jumpto-anywhere command
- [x] fix ref/cite completions (via #956 and the updated analysis)


marlamb:
If `\import` and `\subimport` are added, also the synonyms `\inputfrom` and `\subinputfrom` should be recognized. The same applies to `\includefrom` and `\subincludefrom`, which use `\include` instead of `\input` internally.

PR #991, opened 2016-12-30
Okular viewer prepend pdf path with 'file:' in forward sync, ref #989

### LINKED ISSUE #989: Forward search broken in Okular 1.0
Hi, I recently upgraded to Okular 1.0 (kde apps 16.12)  to find that the forward search does not work anymore when called as usual from LaTeXTools. The problem is discussed here: 
https://bugs.kde.org/show_bug.cgi?id=373855
It seems that a simple syntax fix like in Comment 4 could solve the problem.
Regards,
JG

5chdn:
Issue was introduced with Okular version 0.99.80+, see this tracker: https://bugs.kde.org/show_bug.cgi?id=373855

This PR fixes #989

### DIFF
--- a/viewers/okular_viewer.py
+++ b/viewers/okular_viewer.py
@@ -39,7 +39,7 @@ class OkularViewer(BaseViewer):
     def forward_sync(self, pdf_file, tex_file, line, col, **kwargs):
         self._ensure_okular()
         self._run_okular(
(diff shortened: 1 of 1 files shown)


ig0774:
Alright, I'm satisfied there are no unforeseen consequences. It appears to "just work" on 0.26.1 regardless of the crazy paths I throw at it.

Issue #942, opened 2016-11-15
About the quickpanel


jdumas:
Hi everyone,

I am getting confused with the new quickpanel (`C-r`) to access sections and labels. I have mostly two questions:

1) When I open the "goto anything" panel with `C-r`, I would expect to see only the current sections and paragraph, in the file I am currently editing. If I want to see all of them, I would just use the project-wide search with `^C-R`. It seems to me this behavior is something that has changed recently, can you confirm?

2) When I open the panel with `C-r` to access labels in the project, I need to manually enable the "Show Labels" toggle. But I don't see an option in the config file to show them by default. Did I miss something? Is it possible to do so?

Cheers.

r-stein:
1. Yes there is a new panel to overwrite `C-r` #779. You can disable this behavior in the [settings](https://github.com/SublimeText/LaTeXTools/blob/v3.12.8/LaTeXTools.sublime-settings#L57), but still access this panel via `C-l, C-r`. I personally like that more than the usual `C-r` and `C-shift-r`, because it fits better to the document structure (correct section ordering, only sections inside the document, ...).
2. This is currently not possible, but it would be easy to add a keybinding argument.

jdumas:
1. I like the new panel as well, but I would have liked to have "two" versions of it. One that shows the sections of the current file only (useful to navigate quickly within a file), and one "global" to the project that allows to navigate between the files (and which would override `C-shift-r`).

2. Might I suggest also to provide a "quickpanel" for the references only? E.g. to quickly jump to a figure or a table, I just want to navigate through the list of labels. Right now it's kinda "mixed" together with the sections and I feel that these should be separate entities.

r-stein:
1. You could just disable it to use the default locally and map this to `C-shift-r`:

    ``` js
    {
        "keys": ["ctrl+shift+r"],
        "command": "latex_toc_quickpanel",
        "context":
        [
            { "key": "selector", "operator": "equal", "operand": "text.tex" }
        ]
    },
    ```

2. There is already a function to search for arbitrary commands. Press `ctrl+shift+p` and write `LaTeXTools: Search for commands in document` afterwards provide a comma separated list of commands. This will open a similar navigation panel. You could also bind it to a keybinding and pass commands, e.g.:

    ```js
    {
        "keys": ["ctrl+l", "ctrl+s"],
        "command": "latex_search_command",
        "args": {
            "commands": ["label", "ref", "autoref", "cite"]
        },
        "context":
        [
[...]

PR #1013, opened 2017-01-21
Limit the TOC quickpanel to the current document

r-stein:
This adds a `only_current_file` flag to the toc quickpanel and the command search quickpanel as suggested in #942 by @jdumas.

Open discussion:
We could limit the usual ctrl+r to the file (as usual) and the ctrl+shift+r command to the whole document.

- [x] add the flag to the keybinding (such that the users can see it)

PS. If someone want to try this you can add these keybindings:

``` js
    { 
        "keys": ["ctrl+r"], "command": "latex_toc_quickpanel",
        "args": {"only_current_file": true},
        "context": [
            {"key": "selector", "operator": "equal", "operand": "text.tex"}
        ]
    },
    { 
        "keys": ["ctrl+shift+r"], "command": "latex_toc_quickpanel",
        "args": {"only_current_file": false},
        "context": [
            {"key": "selector", "operator": "equal", "operand": "text.tex"}
        ]
    },
```

### DIFF
--- a/Default (Linux).sublime-keymap
+++ b/Default (Linux).sublime-keymap
@@ -267,14 +267,28 @@ LaTeX Package keymap for Linux
 			{"key": "selector", "operator": "equal", "operand": "text.tex.latex"}],
 		"command": "insert_snippet", "args": {"name":"Packages/LaTeXTools/Text monospace.sublime-snippet"}},
 
-	// Replace the `C-r` overlay with a whole document overlay as opt-out
-	// Also add this overlay to `C-l, C-r`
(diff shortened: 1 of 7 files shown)


r-stein:
I played around with this and would vote for using `ctrl+r` only for the current file and `ctrl+shift+r` for the whole document.

ig0774:
That seems the right thing to do, especially as it's more consistent with the default keybindings for ST. That said, it may be a good idea to add something to the wiki for anyone who's gotten used to the current behaviour and wants to maintain it.

Issue #1011, opened 2017-01-19
Latex fill all freezes ST in large files


msiniscalchi:
This has been happening for at least a couple of months. Whenever latex fill all is invoked, e.g., to complete a reference or citation, or even include a file, ST freezes for a few seconds upon selecting a completion. This is only noticeable in large files, but it is substantial---it happens even on my retina iMac at work, which is no slouch.

The same happens if I use tab completion for a reference or citation.

On the other hand, there is *no* freeze if ref/cite autocompletion is triggered (i.e. when the quick panel is shown as soon as one types `\ref{` etc.). So I suspect there is something related to the fill-all code that causes the freeze, as (I think) ref/cite autocompletion bypasses that.

r-stein:
I think this is because if the [bracket balancing](https://github.com/SublimeText/LaTeXTools/blob/v3.12.15/latex_fill_all.py#L242), i.e. only add the closing bracket if it is not already there. We may set up a threshold for that or only look at a few lines above the insertion point.

msiniscalchi:
Sounds plausible. In addition, bracket balancing doesn't seem to work. I just tried inserting a `\ref{` command. With cwl-completion enabled, I type `\re` and the completion `\ref{key}` is inserted, with `key` selected. I hit Delete, then pres `C-l,C-f`, get the quick panel, select a completion... freeze... and then `\ref{label}}` is inserted, with a double curly brace. So there may well be some issue with capturing the `}` at the time fill-all is invoked.

ig0774:
There's a setting, `smart_bracket_auto_trigger`—and a corresponding toggle—that can be used to disable the bracket balancing, which is probably a good idea to do with a large single-file document, since actually counting brackets is never going to perform well. This also means that it will only add a closing bracket if invoked using the `{` keypress.

Regarding the balancing, it does work, at least in simple cases, e.g.,

![ref](https://cloud.githubusercontent.com/assets/212893/22134679/c69b77ee-dec0-11e6-8031-6662f6a4c645.gif)

But the attempt to take into account the context (i.e., to deal with #766) may not always result in intuitive behaviour in complex situations. In particular, any "unclosed" `{` not in a comment region that occurs before the current fill-all target will result in adding a closing `}`. Again, this goes away if the `smart_bracket_auto_trigger` value is set to `False`.

[...]

PR #1030, opened 2017-02-18
Change the bracket completing algorithm to be less agressive

ig0774:
Per @r-stein's suggestion, the new algorithm defaults to only scanning 5 lines before the selection to determin if there is an unopened bracket that needs to be closed. This, hopefully, helps resolve the issues in #1011.

The number of lines scanned can be configured using the `smart_bracket_look_behind` advanced setting, and the present behaviour of scanning the full document can be restored by using the `smart_bracket_scan_full_document` advanced setting.


ig0774:
Withdrawn. Just realised this doesn't do what I claimed.

Issue #1038, opened 2017-03-08
Why `convert.exe' is unavailable?


ssqqqqqqq:
When checking system, I'm told that `convert.exe' is missing and unavailable. How could this happen?
![9259a5cc7cd98d10bac8ccbd283fb80e7aec90a7](https://cloud.githubusercontent.com/assets/26253417/23693013/0b70170a-040d-11e7-9d3e-c62ce1498203.jpg)

ig0774:
Briefly: that `convert.exe` is the wrong `convert.exe`. Very wrong. The `convert.exe` we're looking for is [the one associated with ImageMagick](https://imagemagick.org/script/convert.php). The `convert.exe` in C:\WINDOWS\System32 is a utility for converting disks to FAT32 format. It's reported as `missing` because when we try to run the version check, the utility returns an error.

Thanks for pointing this out, however. We probably need to add a check to ensure *that* `convert.exe` is never executed, since it could have destructive consequences.

r-stein:
Also convert is not really necessary for the features anymore. We used it to convert pdfs to preview thumbnails to preview included graphics and equations. However we directly use ghostscript (which ships with LaTeX) now.
So as far a I know it is only used to preview svg images. So just ignore, that it is missing. We may inform the user better for what the programs are necessary, but giving good feedback is not that easy.

ssqqqqqqq:
Thanks very much! @ig0774 @r-stein

ig0774:
Fixed in v3.13.0

eric-wieser:
I'm still seeing this behaviour in v3.14.0:

```
Program      Location                                                    Status            Version
-------      --------                                                    ------            -------
sublime      C:\Program Files\Sublime Text 3\sublime_text.exe            available         Sublime Text Build 3126
texify       C:\Program Files (x86)\MiKTeX 2.9\miktex\bin\texify.exe     available         MiKTeX Compiler Driver 2.9.5700 (MiKTeX 2.9)
pdflatex     C:\Program Files (x86)\MiKTeX 2.9\miktex\bin\pdflatex.exe   available         MiKTeX-pdfTeX 2.9.5700 (1.40.16) (MiKTeX 2.9)
xelatex      C:\Program Files (x86)\MiKTeX 2.9\miktex\bin\xelatex.exe    available         MiKTeX-XeTeX 2.9.5700 (0.99992) (MiKTeX 2.9)
[...]

PR #1044, opened 2017-03-13
Don't use convert.exe in the system directory on Windows

ig0774:
Additionally, fixed the system check to NOT show a path when the executable version can't be determined.

This should address (some of) the concerns in #1038.

### DIFF
--- a/st_preview/preview_utils.py
+++ b/st_preview/preview_utils.py
@@ -6,12 +6,16 @@ import traceback
 
 import sublime
 
-try:
+if sublime.platform() == 'windows':
     import winreg
-except:
-    # not on Windows
-    pass
+    import ctypes
+    from ctypes import wintypes
 
+    # wrapper for GetSystemDirectoryW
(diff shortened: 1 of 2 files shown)


st_preview/preview_utils.py
```diff
@@ -33,6 +37,13 @@ def _get_convert_command():
         which('magick', path=texpath) or
         which('convert', path=texpath)
     )
+
+    # DO NOT RUN THE CONVERT COMMAND IN THE SYSTEM ROOT ON WINDOWS
+    if (sublime.platform() == 'windows' and
+            _get_convert_command.result.lower().endswith('convert.exe')):
+        system_root = get_system_root().lower()
+        if _get_convert_command.result.lower().startswith(system_root):
```
eric-wieser: `system_root` is uppercase, so this does not work

eric-wieser: Wait, I take that back

Issue #1043, opened 2017-03-13
Feature Suggestion: Allowing emph/bold/etc. shortcuts to select the word for you. 


diego898:
I've always thought it would be useful for emph, bold, etc. shortcuts to work even though the entire word isn't currently selected (I think thats how Latexing does it). 

So if the cursor is anywhere in a word or on either of its borders, triggering the shortcut would wrap the entire word.

diego898:
Along this line (should I make a separate issue?) It would make sense for environments to also wrap - that is, if I have a bunch of text _selected_, and I then press cmd+L, cmd+E to get a new environment, instead of _replacing_ all the text I had selected, the new environment should _wrap_ all of the text I had selected.

diego898:
Note: currently, I get around this at the word level by first pressing `cmd+d` to highlight the current word (sublime command), and then pressing something like `cmd+l, cmd+e`

ig0774:
I'm not sure I quite understand the environment request. Do you want the selected text in the `\begin{<text>}` and `\end{<text>}`? Surely that already happens. If you want the text *between* the `begin` and `end` commands, then how do you suggest differentiating between text that belongs as arguments to the `begin` and `end` commands and text that belongs in the body?

diego898:
@ig0774 hm - couldnt it jsut be if I have a couple of lines selected and then press `cmd+l,cmd+e` I get my text wrapped with a begin/end environment where the cursor is now in both begin/end so I can change the environment name I just created? Im pretty sure thats how latexing does it

r-stein:
You can wrap text in environments with C-l, C-n

diego898:
:L sorry! Lets just keep it at my original request then. Thanks!

diego898:
@r-stein unless I've missed something, is there currently anyway to have selected a word: `align`, and a keyboard shortcut to turn that into an empty environment?

r-stein:
If your caret is behind the word you can press C-l, e. This should be adapted to also work with selected words.

diego898:
Hm, with my caret behind a word, pressing` c-l,e` just buts a new environment in front of the word, but does not automatically populate the environment name with that word

r-stein:
The cursor needs to be directly behind the word without a space `align|`.

diego898:
[...]

PR #1048, opened 2017-03-16
Implements feature requested in #1043

ig0774:
Adds a command to wrap the selected text in a simple LaTeX command and automatically expands no selection to the current word. This also changes the keybindings from using snippets to using the command for emph, bold, etc. Additionally, I suppose, this makes new keybindings for new similar commands pretty trivial to implement. I've left the snippets in place so they can be used if desired, but obviously they do not have the same word-grabbing rules.


r-stein:
I think this is a little bit to greedy in selecting words and we should insert a snippet if the selection is empty to easy jump-out of it.

ig0774:
The greediness (i.e., selecting the word immediately after or immediately before the cursor) is quite intentional. That was part of the initial request. On the snippet thing: I'm not exactly sure I see inserting, e.g., `\emph{|}` without a tab-out as being that major of an inconvenience. Presumably inserting such a command means you want to add to its content...

r-stein:
> The greediness (i.e., selecting the word immediately after or immediately before the cursor) is quite intentional. That was part of the initial request.

This does not only select words, but also other characters, e.g. if you have `foo bar.|` and press C-l,C-b you will have `foo bar\textbf{.}`.

> On the snippet thing: I'm not exactly sure I see inserting, e.g., `\emph{|}` without a tab-out as being that major of an inconvenience. Presumably inserting such a command means you want to add to its content...

The snippets usually only have a few words/one sentence, so I personally just press it, write the content, and press tab to go further.

ig0774:
> This does not only select words, but also other characters, e.g. if you have foo bar.| and press C-l,C-b you will have foo bar\textbf{.}.

That's a good point and a limitation of the `word()` function. I suppose we can check if it's a one-character string and matches one the `word_separators` and just treat it like an empty string.

> The snippets usually only have a few words/one sentence, so I personally just press it, write the content, and press tab to go further.

Good point. I think I was thinking broadly in terms of getting this working across multiple selections, which is hard to do with snippets, where the selections might vary from one another. The common case, though, is probably just a single cursor and that could be optimised to use a snippet, I suppose.

ig0774:
@r-stein: 1acf8ab implements what I [mentioned above](https://github.com/SublimeText/LaTeXTools/pull/1048#issuecomment-300493543). I'd be greatful if you could take a look.

r-stein:
I would use the expand feature very conservative, because you always can just manually expand it before triggering, but not the other way.
So I would do it like this:
```
if all selections are empty and (preceded by a word character or inside a word):
    wrap the word around the selection with `textbf{....}`
else:
    insert_snippet("\textbf{${1:$SELECTION}}$0")
```

Issue #1056, opened 2017-03-23
On save (Ctrl+S) Updating LatexTools Cache takes a long time


amiroshn15:
When saving a file (Ctrl+S) the text Updating LatexTools Cache [ = ]" appears in the bottom bar  and stays there for about 60-100 seconds. ST3 becomes frozen. I installed the version 3.12.16 to see what happens there. This does not occur. It seems there is an issue with copying cache on save.

Thank you, 
Alexey.

r-stein:
You can disable the caching [in the settings](https://github.com/SublimeText/LaTeXTools/blob/v3.13.2/LaTeXTools.sublime-settings#L606). I think we should make this opt-in, because it slows down Sublime Text on large documents if you save a often.

gerbaulet:
I can confirm that behaviour; "Updating LaTeXTols cache" takes a long time. I would be interested in what has changed in the last update to evoke this?

benruijl:
Same here. I have many large open latex files, and sublime becomes unusable.

r-stein:
I will just cut a release to disable (opt-in) that feature and we can fix the performance issues afterwards. Actually it should increase the performance by moving calculation stuff in the background.

benruijl:
There are perhaps some small things you can do to improve the situation. For example, only update the cache when the saved file is changed, and allow only one cache update at the same time. If I now press Ctrl+S ten times very quickly, I can cripple sublime.

r-stein:
Yes, that is a good idea. I have just released `v3.13.3` to disable it by default and do nothing if it is disabled. Afterwards we can check that we don't start, if the refresh thread is still running and check for changes.

ig0774:
I've implemented the "don't start, if the refresh thread is still running" bit in #1058. I'm not entirely sure how to go about the "check for changes" bit. The problem is that the view is (obviously) never going to be dirty in the `on_post_save` event. It would be possible to check this in the `on_pre_save` event, but we might have to move to using the non-async versions in order to get the timings rights.

r-stein:
@ig0774 we can use the `view.change_count()` for that (just store the change count of the last time we ran it)

ig0774:
Cool!

ig0774:
Ok, #1058 now uses some logic to check for changes. It's implemented in two separate ways since `view.change_count()` doesn't exist on ST2, but since the async methods don't exist either, I'm not concerned about the timing issues with just checking `view.is_dirty()` in `on_pre_save`.

Issue #1054, opened 2017-03-22
No module named multiprocessing


jyuu:
Just updated to 3.13.1 today and haven't been able to compile any .tex file: 

Traceback (most recent call last):
  File "./01_reload_submodules.py", line 101, in <module>
    __import__(mod)
  File "./latextools_utils/__init__.py", line 17, in <module>
    from .settings import get_setting
  File "./latextools_utils/settings.py", line 10, in <module>
    from latextools_utils.utils import run_on_main_thread
  File "./latextools_utils/utils.py", line 11, in <module>
    from multiprocessing import cpu_count
  File ".\multiprocessing\__init__.py", line 83, in <module>
ImportError: No module named _multiprocessing

r-stein:
You are on Sublime Text 2?

jyuu:
Yeah on sublime text 2

jyuu:
Works on sublime 3 though! Thanks

r-stein:
I made a quickfix to at least let in run on ST2. Could you please have a look at it @ig0774

ig0774:
Technically, the number of CPU cores isn't that important. However, it's relatively easily done in Python, as [the default implementation shows](https://github.com/python/cpython/blob/2.6/Lib/multiprocessing/__init__.py#L108), which could remove the dependency on `multiprocessing` which is broken on ST2 on Windows and Linux. Probably should roll this into a future release, but for now, the quickfix is fine.

PR #1058, opened 2017-03-23
Improve cache sanity

ig0774:
An attempt to fix a series of issues raised from the v3.13 release:

* Caches now use a shared, lazy thread pool (no one mentioned this explicitly, but it should stop some resource hogging)
* Saves when the cache for the document is being rebuilt are ignored (as suggested by @benruijl in [#1056](https://github.com/SublimeText/LaTeXTools/issues/1056#issuecomment-288772816)
* `cpu_count()` should work under ST2 (dealing with #1054)
* Updating analysis now invalidates the rest of the local cache, but only after the new analysis is cached
* ThreadPool is properly stopped on plugin unload (i.e. when upgrading or removing LaTeXTools, etc.)
* Addresses several minor bugs in the current implementation

I haven't done anything to remake the cache on save or load the default behaviour.

### DIFF
--- a/latextools_cache_listener.py
+++ b/latextools_cache_listener.py
@@ -59,7 +59,7 @@ class LatextoolsCacheUpdater(object):
         for step in self._steps:
             try:
                 step()
-            except:
+            except Exception:
                 traceback.print_exc()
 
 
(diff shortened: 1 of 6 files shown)


r-stein:
I wouldn't put too much effort to bring features to ST2 as long as it works there. In the future we may also need a cache invalidation chain, because only re-caching analysis does not refresh own-command completion etc..

r-stein:
It seems that caching usually (?) fails on this PR.
Paste in the ST console:
``` py
from LaTeXTools.latextools_utils import cache;
```
and then two times:
``` py
cache.cache("test_root", "test_cache_name", lambda: (print("Cache failed."), [1, 2, 3])[1])
```
Only the first time it should print `Cache failed.`, but it prints it both times on my PC.

ig0774:
Another feature to add: lazy-loading and unloading the cache ThreadPool. Right now it's loaded when the cache is imported and is obviously uselessly using cycles unless LaTeXTools is in use.

r-stein:
But theat feature should be an other PR.

ig0774:
Closing as I can't get the performance of this right. Need a different approach.

Issue #1070, opened 2017-04-04
Get \cite{} and \ref{} completions from open buffers


Aligorith:
When there is no texroot stuff defined, it would be good if the \cite{} and \ref{} commands would just look for the relevant labels in the open buffers (i.e. any .bib files open in the current window/session for \cite{], and any .tex files for \ref{}), instead of just popping up the annoying "No bib files found" error or not allowing any completions at all.

I'm currently working my thesis and have a complicated multi-folder setup that I'm compiling out of Sublime via a batch script.

Aligorith:
I originally opened this as a feature request, but quickly found that it was much easier than I expected to be able to quickly patch this. Now happily using this fix locally :D

PR #1071, opened 2017-04-04
Fix for issue #1070 - Autocomplete for \cite{} and \ref{} looks in open buffers too

Aligorith:
Fixes for  https://github.com/SublimeText/LaTeXTools/issues/1070

### DIFF
--- a/latex_cite_completions.py
+++ b/latex_cite_completions.py
@@ -271,6 +271,23 @@ def find_bib_files(root):
     return cache.LocalCache(root).cache('bib_files', _find_bib_files)
 
 
+def find_open_bib_files(active_window_only=False):
+    # get list of windows we want to check for valid buffers
+    if active_window_only:
+        windows = [sublime.active_window()]
+    else:
+        windows = sublime.windows()
+
+    # find which buffers are .bib files
+    result = []
+    for window in windows:
+        result += [v.file_name() for v in window.views() if v.file_name().endswith(".bib")]
+    print("find_open_bib_files: windows = %d, result = %d" % (len(windows), len(result)))
+
+    # remove duplicates
+    return list(set(result))
+
+
 def run_plugin_command(command, *args, **kwargs):
     '''
     This function is intended to run a command against a user-configurable list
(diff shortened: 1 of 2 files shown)


latex_cite_completions.py
```diff
@@ -271,6 +271,23 @@ def _find_bib_files():
...
+    # get list of windows we want to check for valid buffers
+    if active_window_only:
+        windows = [sublime.active_window()]
+    else:
+        windows = sublime.windows()
+
+    # find which buffers are .bib files
+    result = []
+    for window in windows:
+        result += [v.file_name() for v in window.views() if v.file_name().endswith(".bib")]
```
r-stein: Change this to `if v.file_name() is not None and v.file_name().endswith(".bib")` or just to `if (v.file_name() or "").endswith(".bib")` to avoid an error, the view is not saved.

latex_ref_completions.py
```diff
@@ -55,14 +55,29 @@ def find_labels_in_files(root, labels):
         labels.append(command.args)
 
 
+# grab labels from all open buffers too
+def find_labels_in_open_files(views, labels):
+    for view in views:
+        # XXX: Hardcoded to only look for tex files, so that it doesn't try 
+        # looking for label commands in support files
+        if view.file_name().endswith(".tex"):
```
r-stein: Same as above

r-stein:
I am not whether this is really an improvement, because the completions should depend on the tex document and not on the open files. However we may make it opt-in as a "hack" for people, who use a setup, which is currently not supported.

latex_cite_completions.py
Aligorith: Done :)

latex_ref_completions.py
```diff
@@ -55,14 +55,29 @@ def find_labels_in_files(root, labels):
         labels.append(command.args)
 
 
+# grab labels from all open buffers too
+def find_labels_in_open_files(views, labels):
+    for view in views:
+        # XXX: Hardcoded to only look for tex files, so that it doesn't try 
+        # looking for label commands in support files
+        if view.file_name() is None or view.file_name().endswith(".tex"):
```
ig0774: This could probably be more sensibly implemented by checking the syntax (something like `view.score_selector('text.tex.latex', 0) > 0` or at least checking for various extensions supported using the `is_tex_file()` function in the `latextools_utils.is_tex_file` module.

latex_cite_completions.py
```diff
@@ -271,6 +271,22 @@ def _find_bib_files():
     return cache.LocalCache(root).cache('bib_files', _find_bib_files)
 
 
+def find_open_bib_files(active_window_only=False):
```
ig0774: I'd prefer that the default option was to search the active window only since its possible to have multiple unrelated TeX documents open at the same time.

PR #1080, opened 2017-04-15
Added "recommended packages" to the documentation

r-stein:
This adds a "recommended packages". I think it is useful to have this section, because not every user may be super confident with all Sublime Text packages and it will be helpful for new users to have packages, which are working with LaTeX. In addition it may give some configuration ideas to all users.


ig0774:
I like this idea. One thing though: should we make
It clear that it's packages for Sublime rather than, say, LaTeX?

r-stein:
Yes, I will change that to make it clear.

PR #1111, opened 2017-05-11
Popup for quick and easy insertion of symbols inside math environments

r-stein:
This adds a popup, which can be used to insert symbols and operators inside math environments. The basic idea is to use an input method, which is as intuitive as possible, but is still useful for advanced users, who just want it to be fast. You press the trigger key and then write the trigger for the corresponding symbol. The trigger tries to emulate the appearance of the resulting symbol, e.g. `a` for `α`, `->` for `→`, `-->` for `⟶`, and `|->` for `↦`.
If the prefix only matches one remaining symbol it will be inserted automatically otherwise you use up/down to select one and tab/enter/space to insert it.

The triggers and corresponding commands are written in a JSON file and user configureable. Even after changes the preview (2nd column) should also be good, because I wrote a small LaTeX to html converter, which knows a lot of symbols (by a yaml meta data file) and understands the basic commands.

Add this to your keymap to create a trigger key (change the key to the one you wish to use):

``` js
    {
        "keys": ["#"],
        "command": "latextools_math_popup_show",
        "context":
        [
            { "key": "selector", "operator": "equal", "operand": "text.tex.latex meta.environment.math" },
        ]
    },
```

*(The image problems in this gif are due to the gif record program, not ST or this plugin)*
insert_math_popup

*This is still a beta version and I will need to change several things, before we can release it*

- [ ] customizable CSS + CSS for bright schemes


msiniscalchi:
Great idea @r-stein ! 

This is something I also considered a while back. TeXmacs has a similar mechanism, except that (i) substitution happens automatically, so e.g. you get a `\rightarrow` as soon as you type `->`, and (ii) to enter "variants" [e.g. `\longrightarrow` instead of `\rightarrow`, or `\succcurlyeq` instead of `\geq`] one types a `*`, which can possibly be repeated.

The implementation would be quite different from what you have---I'm just mentioning this as something to keep in mind. The main user-facing difference is that TeXmacs is completely WYSIWYG, so you actually get a nice, typeset arrow when you enter `->`. So what works for TeXmacs need not work for LaTeXTools.

Issue #1115, opened 2017-05-16
\cite autocompletion not working anymore


musm:
```
2017-05-16 15:01:56,839: 14060: DEBUG: on_selection_modified
Traceback (most recent call last):
  File "C:\Program Files\Sublime Text 3\sublime_plugin.py", line 812, in run_
    return self.run(edit, **args)
  File "C:\Users\Mus\AppData\Roaming\Sublime Text 3\Packages\LaTeXTools\latex_fill_all.py", line 1073, in run
    reraise(*sys.exc_info())
  File "C:\Users\Mus\AppData\Roaming\Sublime Text 3\Packages\LaTeXTools\latex_fill_all.py", line 41, in reraise
    raise value
  File "C:\Users\Mus\AppData\Roaming\Sublime Text 3\Packages\LaTeXTools\latex_fill_all.py", line 1067, in run
    view, prefix, line[::-1]
  File "C:\Users\Mus\AppData\Roaming\Sublime Text 3\Packages\LaTeXTools\latex_cite_completions.py", line 474, in get_completions
    completions = get_cite_completions(view)
  File "C:\Users\Mus\AppData\Roaming\Sublime Text 3\Packages\LaTeXTools\latex_cite_completions.py", line 404, in get_cite_completions
    completions = run_plugin_command('get_entries', *bib_files)
TypeError: run_plugin_command() argument after * must be a sequence, not object
2017-05-16 15:01:56,842: 14060: DEBUG: on_post_text_command
2017-05-16 15:01:56,945: 14060: DEBUG: on_idle
2017-05-16 15:02:04,851: 14060: DEBUG: on_hover
2017-05-16 15:02:05,901: 14060: DEBUG: on_hover
```

`\ref` works but for some reason `\cite` no longer works?

r-stein:
Can you delete your bibcache? (C-shift-p "LaTeXTools: Clear bibliography cache) We are aware of this problem and working on it #1081

musm:
```
['C:\\Users\\Mus\\Dropbox (MIT)\\publications\\phd_thesis\\working_copy\\main.bib']
2017-05-16 15:17:04,336: 14972: DEBUG: on_modified
2017-05-16 15:17:04,337: 14972: DEBUG: post_on_modified
2017-05-16 15:17:04,339: 14972: DEBUG: on_selection_modified
Traceback (most recent call last):
  File "C:\Program Files\Sublime Text 3\sublime_plugin.py", line 812, in run_
[...]

PR #1117, opened 2017-05-16
Extend parser to allow more citekey names

ig0774:
This should fix the underlying issue in #1115.


musm:
thank you @ig0774 I will test this and report back when it is available on the prerelease 

love all the work here keep up

PR #1120, opened 2017-05-18
Rewrite the toggle settings command to unify them in a file

r-stein:
In addition the "show toggle keys" dialog is opened if the key is unknown. This dialog contains information about the trigger key sequence to toggle each setting.


r-stein:
I will add a "description" entry, such that you can change the title in the overlay.

r-stein:
Now it looks like this (3rd line is description):

example

PR #1131, opened 2017-05-23
Add a command to automatically insert label contents

r-stein:
This creates a command to automatically create and insert labels in a smart way. It supports both single and multiple selections. The content of the label is auto-detected based on the line above the insertion point and the surrounding environment.

auto_label_example

TODO:
- [x] move meta information to settings
- [x] enable/disable via settings similar to fill-all helper


LaTeXTools (Advanced).sublime-settings
```diff
@@ -39,6 +39,39 @@
 
 
 // ------------------------------------------------------------------
+// Auto-label config
```
r-stein: It would be good if someone could review these settings. Not sure if there is some standard for this. For example I have seen both "ssub" and "sss" for subsection.

PR #1152, opened 2017-06-11
Added a setting to define an additional bib file in the settings

### LINKED ISSUE #472: Setting path for the bib file location
Hello and first of all thank you for this amazing tool!!

I don't know if it is possible but I wanted to ask whether there is a way where I could set the path to where the bib file is located in the settings file such that when I type \cite{} it loads the references from that bib file automatically, because currently my bib file is not located in the same directory as the .tex files.

Thanks again!! :+1:

r-stein:
This fixes #472

### DIFF
--- a/latex_cite_completions.py
+++ b/latex_cite_completions.py
@@ -280,10 +280,10 @@ def find_bib_files(root):
     additional_file = get_setting("additional_bibliography_file")
     if additional_file:
         def _make_abs_path(file_path):
-            root_folder, _ = os.path.split(root)
             if not file_path.endswith(".bib"):
                 file_path += ".bib"
             if not os.path.isabs(file_path):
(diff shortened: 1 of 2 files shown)


latex_cite_completions.py
```diff
@@ -276,6 +276,21 @@ def _find_bib_files():
             cache.LocalCache(root).set('bib_files', result)
         except:
             pass
+    # if a an additional_file is set append it to the result
+    additional_file = get_setting("additional_bibliography_file")
+    if additional_file:
+        def _make_abs_path(file_path):
+            root_folder, _ = os.path.split(root)
```
PeterPablo: For clarity this line might be moved into the `isabs` check below as it is the only consumer of `root_folder`.

PR #1155, opened 2017-06-12
Added a command to smart paste images

### LINKED ISSUE #1154: A command to paste a image from clipboard to the code?
On the question [Drag and drop image files to LyX](https://tex.stackexchange.com/questions/58673/drag-and-drop-image-files-to-lyx) they show:

> Not by drag & drop, but it is possible with copy & paste. Just put the image in the clipboard and select Edit->Paste Special. LyX will then immediately open a "Save as" dialog, where you can specify where and under which name the image will be saved. However, you can also just press Enter to choose the selected name and path.

A plugin like this for Sublime Text would help a lot when [writing latex code](https://tex.stackexchange.com/questions/19176/how-to-insert-an-image-into-latex-document).

When the command is issue, it would:

1. Take the contents of the clipboard
1. Check whether it is an image.
1. Ask the user for a folder where to save the image, already suggesting the last used path and a name based on some information it can get as the current date and time.
1. Insert the image path on the cursor current position as:
```latex
\begin{figure}
\centering
        \includegraphics[totalheight=8cm]{images/verticalcell.eps}
    \caption{used by \citet[p.~4]{XXXXX}.}
    \label{fig:verticalcell}
\end{figure}
```

Originally posted on: [$28991](https://forum.sublimetext.com/t/a-command-to-paste-a-image-from-clipboard-to-the-code/28991) 

Related issues:

1. [sublime-evernote$24](https://github.com/bordaigorl/sublime-evernote/issues/24) Insert Image?
1. [forum$1965](https://forum.sublimetext.com/t/drag-and-drop-image-to-code/1965) Drag and drop image to code

r-stein:
This downloads the images from the internet and creates the figure environment to include the image into the document.

This resolves #1154 suggested by @evandrocoan

smart_paste_image_example

TODO:

- [x] setting to disable the smart paste
- [x] move the template to the setting
- [x] support for `\graphicspath`
- [x] add documentation in the docs


evandrocoan:
Can sublime request a image from clipboard?
I found this topic: https://forum.sublimetext.com/t/sublime-get-clipboard-return-file-path/12294

I put a screenshot on the clipboard and then I tried to do that with: 
```
        clipboard = sublime.get_clipboard().strip()
        print( "clipboard: " + str( clipboard ) )
```
But I get:
```
clipboard: 
Unable to extract text from the clipboard, available formats: CF_BITMAP, CF_DIB, CF_DIBV5
```

r-stein:
I understood the request to paste URL and download the images, but to copy and paste images from the file manager also seems resonable. However that is OS specific and hence more complicated.

r-stein:
In addition we can't install pip libraries and can't use `PIL` oder `pygtk` as in the solutions.

evandrocoan:
So it seems a challenge. Could we request support from Sublime Text core for this feature?

Also, I just noticed I clipboard program can save the images to the file system, and automatically copy their path to the clipboard as `D:\User\Downloads\envinronmentHead.png` when they are saved. Then can it "download" the image from the own computer file system?

On this case, it would already suggest the own picture file name. This could be used also from the web pictures.

r-stein:
Using the path like this is trivial and I was looking forward to add this. (Btw. you can also copy the path on windows by shift+rightclick on the file and select the entry "Copy as path" in the context menu)

I would just insert the name if it is a subfolder of the root/graphicspath folder and copy the file as in the url case otherwise.

evandrocoan:
> Btw. you can also copy the path on windows by shift+rightclick on the file and select the entry "Copy as path" in the context menu
 
I didn't know that, thanks!

> I would just insert the name if it is a subfolder of the root/graphicspath folder and copy the file as in the url case otherwise.

On cases I am inserting pictures from folder like `D:\User\Downloads\envinronmentHead.png`, I would like it to suggest the original picture name `envinronmentHead.png` when importing them to the project.

On cases it cannot guess a name I would like it to remember the last typed name for the last saved picture. This is very useful when I am working on a document and inserting images from the clipboard or elsewhere, as the system clipboard.

r-stein:
I have added an option to paste file paths for offline images.
- If the image is in a subfolder of the tex root it will insert the relative name
- If the image is somewhere else it will ask for a name and copy the image as in the URL case

evandrocoan:
Thanks for the efforts. Now the support seems very nice for images. I think this is good for a pull request.
The other feature about hooking the clipboard could be for another time. These improvements are already very helpful.

Issue #1161, opened 2017-06-13
Cannot determine completion type for current selection, for multi cursors


evandrocoan:
I created this command I ran using the [Chain Of Command](https://github.com/jisaacks/ChainOfCommand) package on my user folder, based on:

1. https://github.com/SublimeText/LaTeXTools/issues/1156 Ability to insert project already existent images command by a quick panel

(User/Commands/LaTeX.sublime-commands)
```javascript
[
	{
		"caption": "LaTeXTools: Insert Image",
		"command": "chain",
		"args": {
		    "commands": [
		        ["insert_snippet", {"contents": "\\begin{figure}\n\\centering\n\\includegraphics[scale=1.0]{$TM_SELECTED_TEXT$1}\n\\caption{$1}\n\\label{fig:$1}\n\\end{figure}"}],
		        ["latex_fill_all"]
		    ]
		},
	},
]
```
This creates a selection like this:

![image](https://user-images.githubusercontent.com/5332158/27099832-945a6782-5052-11e7-914e-f0fae9bdfe84.png)

However when there are multiple cursors, the Sublime Text console receives a:
```
Cannot determine completion type for current selection
```
When I run the command `LaTeXTools: Insert Image`.

___

Related issues:

1. https://github.com/SublimeText/LaTeXTools/issues/815 Cite-autocompletion no longer working

r-stein:
Yeah you need to use `$1` only inside includegraphics and $2 inside the other ones.

evandrocoan:
But I want to it to autocomplete the image name/path on all these fields. Currently I can do this with the smart paste feature, setting it to `<<image_name>>` 3 times at `User\LaTeXTools (Advanced).sublime-settings`:
```javascript
{
    "smart_paste_image_snippet": [
        "\\begin{figure}[H]",
        "\\centering",
        "\\includegraphics[scale=1.0$2]{<<image_name>>}",
        "\\caption{<<image_name>>$TM_SELECTED_TEXT$1}",
        "\\label{fig:<<image_name>>${1/([0-9a-zA-Z_])|([^0-9a-zA-Z_]+)/\\L\\1(?2:_:)/g}}",
        "\\end{figure}",
    ],
}
```

ig0774:
You could try changing the `"latex_fill_all"` to something like `"latex_fill_all", {"completion_type": "input", "force": false}`, which may work (though I can't make any guarantees... the FillAll bit wasn't really designed to accommodate this sort of thing).

[...]

PR #1162, opened 2017-06-14
Created the parameter single_selection for latex_fill_all

evandrocoan:
command. It is a boolean indicating whether or not to ignore
multiple cursors/carets on the text. This is useful when creating
a snippet which defaults to creating multiple carets/cursors and
want to force the completion for some specific scope as `input`
or `cite`. Therefore we must to ignore the multi selection,
otherwise it will ignore the input scope specified.

Solves the issue:
1. https://github.com/SublimeText/LaTeXTools/issues/1161 Cannot determine completion type for current selection, for multi cursors

For that solution, you just need to use this command on the
command palette, adding the `single_selection` and `force` arguments:
```latex
[
	{
		"caption": "LaTeXTools: Insert Image",
		"command": "chain",
		"args": {
		    "commands": [
		        ["insert_snippet", {"contents": "\\begin{figure}\n\\centering\n\\includegraphics[scale=1.0]{$TM_SELECTED_TEXT$1}\n\\caption{$1}\n\\label{fig:$1}\n\\end{figure}"}],
		        ["latex_fill_all", {"completion_type": "input", "force": true, "single_selection" : 0} ]
		    ]
		},
	},
]
```

### DIFF
--- a/latex_fill_all.py
+++ b/latex_fill_all.py
@@ -876,11 +876,15 @@ class LatexFillAllCommand(
         is disabled
 
     :param single_selection:
-        boolean indicating whether or not to ignore multiple cursors/carets on
-        the text. This is useful when creating a snippet which defaults to
-        creating multiple carets/cursors and want to force the completion for
-        some specific scope as `input` or `cite`. Therefore we must to ignore
-        the multi selection, otherwise it will ignore the input scope specified
+        positive integer indicating to ignore multiple cursors/carets on the
+        text and use one specific selection point. This is useful when creating
+        a snippet which defaults to creating multiple carets/cursors and want to
+        force the completion for some specific scope as `input` or `cite`.
+        Therefore we must to ignore the multi selection, otherwise it will
+        ignore the input scope specified
+        the accepted values start on 0 for the first selection, 1 for the second
+        and etc. They must to be lower then the maximum selections, otherwise
+        this setting will be ignored.
     '''
 
     def is_visible(self, *args):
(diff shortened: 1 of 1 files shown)


ig0774:
Instead of this, which only works if the snippet being matched is the first selection, I'd prefer a more generic solution where you could indicate, e.g., `single_selection = 1` for the first selection, `single_selection = 2` for the second, etc. This would enable this to be used for snippets where the matching selection is not necessarily the first in the selection.

evandrocoan:
Thanks, I did not had though of that. I added it on the last commit.

latex_fill_all.py
```diff
@@ -874,22 +874,46 @@ class LatexFillAllCommand(
...
+        this setting will be ignored.
     '''
 
     def is_visible(self, *args):
         return bool(self.view.score_selector(0, "text.tex.latex"))
 
     def run(
         self, edit, completion_type=None, insert_char="", overwrite=False,
-        force=False
+        force=False, single_selection=-1
```
r-stein: I would prefer `single_selection=None` (and `single_selection is not None and ...` later)

latex_fill_all.py
```diff
@@ -874,22 +874,46 @@ class LatexFillAllCommand(
...
-        for sel in view.sel():
-            point = sel.b
-            if not view.score_selector(point, "text.tex.latex"):
-                self.complete_brackets(view, edit, insert_char)
-                return
+        selections = view.sel()
+        selections_len = len(selections)
+
+        # initialize the `point` variable to a proper value, to be used later
+        if selections_len:
```
r-stein: Please flatten this by
```py
if not selections_len:
    return
```

evandrocoan:
I applied the changes, but I am not sure about using the double negative `is not None` (!!var).  Why would be necessary to use `if single_selection is not None and ...`, instead of only `if single_selection and ...`?

r-stein:
if `single_selection` is `0` it would have the wrong behavior.

evandrocoan:
Thanks, I fixed it shifting the selection positions from starting on 0 to 1. I did so because if we do `is not None` we would be negating it on every call to the command `latex_fill_all`. Now it is only doing a extra call when the parameter `single_selection` is actually used, which should happens almost never.

PR #1179, opened 2017-07-19
Math preview: Add current tex dir to search path

nanoant:
Adding current tex root dir to `TEXINPUTS` search path allows
`"preview_math_template_preamble"` to contain `\input` or `\usepackage` that pulls
files relatively to edited document, rather than temporary folder, e.g.

    "preview_math_template_preamble": "\\input{symbols.tex}"

### DIFF
--- a/st_preview/preview_math.py
+++ b/st_preview/preview_math.py
@@ -14,6 +14,7 @@ import sublime_plugin
 
 from ..parseTeXlog import parse_tex_log
 
+from ..getTeXRoot import get_tex_root
 from ..latextools_utils import cache, get_setting
 from ..latextools_utils.external_command import execute_command
 from . import preview_utils
@@ -110,7 +111,7 @@ def plugin_unloaded():
     _lt_settings.clear_on_change("lt_preview_math_main")
 
 
-def _create_image(latex_program, latex_document, base_name, color,
+def _create_image(latex_program, latex_document, base_name, color, tex_root,
                   **kwargs):
     """Create an image for a latex document."""
     rel_source_path = base_name + ".tex"
@@ -126,10 +127,12 @@ def _create_image(latex_program, latex_document, base_name, color,
     with open(source_path, "w", encoding="utf-8") as f:
         f.write(latex_document)
 
+    env = {"TEXINPUTS": os.pathsep.join(['.', os.path.dirname(tex_root), ''])}
+
     # compile the latex document to a pdf
     execute_command([
         latex_program, '-interaction=nonstopmode', rel_source_path
-    ], cwd=temp_path)
+    ], cwd=temp_path, env=env)
 
     pdf_exists = os.path.exists(pdf_path)
     if not pdf_exists:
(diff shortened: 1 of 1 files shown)


r-stein:
Sounds interesting. My main problem with this is, that using `\input` in that preamble is problematic* and should be avoided, so adding a feature to support it is in general not something I am looking forward to.

*Including other files is problematic, because the unique image identifier is build over the corresponding file content, so you will get a the same image for the same equation and already written equations are cached. If you use `\input{otherfile}` and change the content of the other file the file content will stay the same and hence you will have a wrong image and this will lead to problems.
However I will think about this (since it has been requested several times) and may come to a satisfying solution for it.

nanoant:
@r-stein Wow, thanks for instant reply. Indeed this may be not a perfect solution, but I needed it quickly. (La)TeX is pretty arcane and really terse in terms of available command line options and I had no better idea than using this `TEXINPUTS` variable.
But I guess you may have much better idea how to get this implemented well, so I look forward to it.
Cheers, Adam.

r-stein:
I like your solution =), but my problem is more on a meta level "How do I know that the included files have changed?", because we currently only use the file content to cache the document->image relationship [here](https://github.com/SublimeText/LaTeXTools/blob/st3-3.14.1/st_preview/preview_math.py#L723) and this does not work with `\input`.

My ideas would be to search for `\input{...}` in the document and either
- read the last change timestamp of the corresponding file and include it in the ID
- just inline all included documents

r-stein:
Aside: You may state that you agree to an upcoming license #1175, thanks

nanoant:
@r-stein Oh yes, I hereby do agree to release my contribution to LaTeXTools under MIT license.

bordaigorl:
I cannot believe I just created a PR that is identical to this one 😮 #1401

nanoant:
And here we go again 2 years later. Can we have this integrated? I know it is not perfect, but it better than nothing at all. And having predefined macros for symbols is really common case across many projects.

r-stein:
My comment still holds true. I understand the desire for this feature and also want to include it, but it is not easily possible with the current concept and may lead to performance issues if implemented in a different way.
I would be pleased if more/any development was done in LaTeXTools, because I think it's a nice package with a lot of potential for more features (especially with the new sublime text apis).

bordaigorl:
@r-stein I really don't understand why you arw hesitating in merging this.
It seems that your foremost priority is that you can catch *every* change in the files so you know when to recompile. This is nice to have but definitely not essential:

 1. one simply needs to know that included/inputed files are not watched for changes and clear the cache when that happens
 2. the macro files are usually very stable while the formulas change a lot when you need the preview, so this is a perfectly ergonomic solution

On the other hand, not having this facility is a real pain, and the limitation in capabilities really does not justify the insistence on "watchability" of the changes.
Plus, this feature is completely optional and if one prefers to keep everything watchable, then one can simply not use inputs in the settings.

What is the problem???

r-stein:
I think it would be a compromise to merge it as opt-in with the remark that one needs to manually clear the cache in such cases?

nanoant:
@r-stein Just to be clear we are talking about manually running *"LaTeXTools: Clear Cache"* when the `file.tex` from `\input{file.tex}` changes? There is nothing else needed.

Shall I put the remark, or you can do it after merge?

r-stein:
Yes I mean the clear cache command. I will recheck this and create a setting for it

PR #1192, opened 2017-08-08
Fix cache returning unidentified `object()`

ig0774:
Basically, this re-implements `_invalid_object` (the marker object used to mark an invalidated cache entry) in a way that will survive LaTeXTools being reloaded. This is the root cause of a number of issues with bibliography entries returned from the cache being incorrect. While it will work for future caches, current caches will need to be deleted to see any sort of effect.


r-stein:
Aside from this PR it may be useful to give an option for a pure memory cache. For example it does not make to much sense to store the analysis on disk, because it is outdated anyway if you start an other ST session.

Issue #1193, opened 2017-08-09
Auto-Completes for \{, \[, \(, etc. 


lirm-math:
For the current version, the  brackets and quotes 
`left-round-bracket, left-square-bracket, left-curley-bracket, grave, quote, double-quote`
are auto-closed by 
`right-round-bracket, right-square-bracket, right-curley-bracket,quote, quote, double-quote`.
This is a helpful feature. 

But, IMHO, when the `left-brackets` are after a `back-slash`, they are suppoed to be closed by
`right-brackets` after a `back-slash` . Say, `\[\]`. 
And  `backslash-quotes` are used to denote accents (e.g., backslash-dobule-quote-o stands for  ö),
they are not supposed to be ''closed''. 

So, it is better to change the default behavior? 
 
Thank you very much.

r-stein:
Yes we should finally add this (#963 and #917)

Issue #963, opened 2016-11-27
Auto-pair escaped curly braces


FichteFoll:
Here is what I added to my key bindings just now:

```
  // Auto-pair curly brackets in tex
  { "keys": ["{"], "command": "insert_snippet", "args": {"contents": "{$0\\\\}"}, "context":
    [
      { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
      { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
      { "key": "preceding_text", "operator": "regex_contains", "operand": "\\\\$", "match_all": true },
      { "key": "following_text", "operator": "regex_contains", "operand": "^(\\s|\\)|]|\\}|\\\\|$)", "match_all": true },
      { "key": "selector", "operand": "text.tex", "match_all": true }
    ]
  },
  { "keys": ["backspace"], "command": "run_macro_file", "args": {"file": "res://Packages/User/Delete Left Right Twice.sublime-macro"}, "context":
    [
      { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
      { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
      { "key": "preceding_text", "operator": "regex_contains", "operand": "\\\\\\{$", "match_all": true },
      { "key": "following_text", "operator": "regex_contains", "operand": "^\\\\\\}", "match_all": true },
      { "key": "selector", "operand": "text.tex", "match_all": true }
    ]
  },
```

I believe this would be a good addition to LaTeXTools.
Note that I didn't add a binding to skip over the closing brace because ... I don't think there is a good user pattern for it.

r-stein:
Related #917. Yes, we should add that.

randy3k:
A number of similar shortcuts are defined in [LaTeXZeta](https://github.com/randy3k/LaTeXZeta), perhaps we could merge some of them to LaTeXTools if you think that they are beneficial.

r-stein:
LaTeXZeta looks like a really nice addition (I have installed it) and we should use some of those keybinding =)
I am not sure whether we should add a new Default.sublime-keymap (less work/maintaing) file or use the Default ($platform).sublime-keymap (visible to the user).

We may not include the prefix key stuff, because the prefix key `` ` `` doesn't work on german keyboards and I would wish to have something more configurable.

[...]

Issue #917, opened 2016-11-03
`\( ... \)` Math environment autocomplete


skulumani:
Currently it seems that the math environment `$ ... $` will automatically create the matching end notation. 

However, this seems to not work for the LaTeX equivalent `\( ... \)` or `\[ ... \]`. In addition, the [Wikibook](https://en.wikibooks.org/wiki/LaTeX/Mathematics#Mathematics_environments) says that the older notation should be avoided. 

I think the relevant code is defined [here](https://github.com/SublimeText/LaTeXTools/blob/96458b978ab1cdf80e47e4ba9119a0c95a5a7401/Default%20(Linux).sublime-keymap#L289-L309) but I haven't had a chance to fully figure out Sublime/LaTeXTools

r-stein:
Just add:

``` js

    {
        "keys": ["["], "command": "insert_snippet", "args": {"contents": "[$0\\\\]"},
        "context": [
            { "key": "selector", "operator": "equal", "operand": "text.tex.latex" },
            { "key": "preceding_text", "operator": "regex_contains", "operand": "\\\\$", "match_all": true },
        ]
    },
    {
        "keys": ["("], "command": "insert_snippet", "args": {"contents": "($0\\\\)"},
        "context": [
            { "key": "selector", "operator": "equal", "operand": "text.tex.latex" },
            { "key": "preceding_text", "operator": "regex_contains", "operand": "\\\\$", "match_all": true },
        ]
    },
```

to your keymap. In addition only `$$...$$` should be avoided.

skulumani:
That's great. Thanks for the help!

ig0774:
@r-stein: Might these keybindings be worth including in the package?

r-stein:
Yes, we could also include other nice keybindings like

``` js

    {
        "keys": ["^", "^"], "command": "insert_snippet", "args": {"contents": "^{${1:$SELECTION}}$0"},
        "context": [
            { "key": "selector", "operator": "equal", "operand": "text.tex.latex meta.environment.math, text.tex.latex string.other.math" },
        ]
    },
    {
        "keys": ["_", "_"], "command": "insert_snippet", "args": {"contents": "_{${1:$SELECTION}}$0"},
        "context": [
            { "key": "selector", "operator": "equal", "operand": "text.tex.latex meta.environment.math, text.tex.latex string.other.math" },
        ]
    },
```

Also auto insert \right in a similar way, if \left is present would be useful.

[...]

PR #1194, opened 2017-08-09
Auto-pair escaped braces, brackets, and parens

r-stein:
This has been suggested by @Romeoly in #1193, by @skulumani in #917, and by @FichteFoll in #963.

It currently does:
(&#124; is the caret position)

insert | get
--- | ---
`\{` | `\{`&#124;`\}`
`\[` | `\[`&#124;`\]`
`\(` | `\(`&#124;`\)`
`\left(` | `\left(`&#124;`\right)`

It also changes the backpace behavior (same for parens and braces):

before pressing | after pressing
--- | ---
`\[`&#124;`\]` | &#124;
`\[\]`&#124; | `\[`&#124;
`\[`&#124; | &#124;
`\left(`&#124;`\right)` | &#124;

I am open to discussion for more worthy additions.

### DIFF
--- a/Default.sublime-keymap
+++ b/Default.sublime-keymap
@@ -10,6 +10,15 @@
     },
 
     // Auto-pair escaped curly braces/brackets/parens
+    // improve enter key behavior
+    {
+        "keys": ["enter"], "command": "insert_snippet", "args": {"contents": "\n\t$0\n"},
+        "context": [
+            { "key": "selector", "operator": "equal", "operand": "text.tex.latex" },
+            { "key": "preceding_text", "operator": "regex_contains", "operand": "(?:\\\\left)?(?:\\\\)?(\\(|\\[|{)$", "match_all": true },
+            { "key": "following_text", "operator": "regex_contains", "operand": "^(?:\\\\right)?(?:\\\\)?(\\)|\\]|})", "match_all": true },
+        ]
+    },
     {
         "keys": ["{"],
         "command": "insert_snippet",
@@ -19,8 +28,8 @@
             { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
             { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
             { "key": "preceding_text", "operator": "regex_contains", "operand": "\\\\$", "match_all": true },
-            { "key": "following_text", "operator": "regex_contains", "operand": "^(\\s|\\)|]|\\}|\\\\|$)", "match_all": true },
-            { "key": "selector", "operand": "text.tex", "match_all": true }
+            { "key": "following_text", "operator": "regex_contains", "operand": "^(\\s|\\)|]|\\}|\\\\|\\$|$)", "match_all": true },
+            { "key": "selector", "operand": "text.tex.latex", "match_all": true }
         ]
     },
     {
@@ -32,8 +41,8 @@
             { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
             { "key": "selection_empty", "operator": "equal", "operand": true, "match_all": true },
             { "key": "preceding_text", "operator": "regex_contains", "operand": "\\\\$", "match_all": true },
-            { "key": "following_text", "operator": "regex_contains", "operand": "^(\\s|\\)|]|\\}|\\\\|$)", "match_all": true },
-            { "key": "selector", "operand": "text.tex", "match_all": true }
+            { "key": "following_text", "operator": "regex_contains", "operand": "^(\\s|\\)|]|\\}|\\\\|\\$|$)", "match_all": true },
+            { "key": "selector", "operand": "text.tex.latex", "match_all": true }
         ]
     },
     {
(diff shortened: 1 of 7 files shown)


JackGin:
>I am open to discussion for more worthy additions.
Obviously, \left( should be also auto-paired with \right), with shortcuts similar to LaTeXYZ's like (( -> \left( and so on.

r-stein:
@JackGin 
> Obviously, \left( should be also auto-paired with \right)
Which should be paired?
- ( and )
- \{ and \}
- [ and ]

JackGin:
Each one, imho. And \langle with \rangle, too.

Romeoly:
I can remember the following, 

- \lceil, \rceil,      % the ceiling function, 
- \lfloor, \rfloor, % the floor function, 
- \|, \|,                % ||x||, the norm of x.
- |, |                   % |x|,  the absolute value of x,
- \lvert, \rvert    % alias of    |,   |,
- \lVert, \rVert    % alias of  \|,  \|,
- \vert, \vert,      % alias of   |,   |, 
- \Vert, \Vert,      % alias of \|,  \|,
 
AFAIR, the < and > in LaTeX are not recommended to use as brackets (unless in beamer-related documents),
and so, I don't think any < should be auto-closed by a >.  In fact, the correct command for angle-brackets is 
- \langle, \rangle, 

I don't think  '(quote) or "(double-quote) should be closed either, because the right command for
quotes are 
- `single-quote', and 
- ``double-quote''.
 
Graves are left-quotes and quotes are right-quots. The double quotes are merely two signle-quotes.

What's more, if we are not restricted to brackets and quotes, I'd like to point out that the other common pairing commands (for maybe package-writers) include 
- \bgroup, \egroup,
- \begingroup, \endgroup, and
- \csname, \endcsname. 

The command \verb is among the most strange commands. It is closed by the first char (which must be non-letter and non-space) directly are itself. 
For example, 
- \verb"how do you do?"  % is the same as 
- \verb(how do you do(    % as
- \verb)how do you do)    % as
- \verb{how do you do{    % as
- \vreb}how do you do}    % as
- \verb!how do you do!    % etc

There are some exceptions, 
- \catcode, and 
- \string. 

When Grave appears after \catcode, it not to be closed by '. 
And \string  turns the token after it into a string. For example, \string\langle prints seven chars {\, a, n, g, l, e} to the document instead of the \langle-char.  Therefore,  I don't think anything following \string should be closed. 

I will continue updating should I think of more.

r-stein:
@Romeoly Thanks for the list, but what would you expect LaTeXTools to do in that cases?

Romeoly:
@r-stein I thought braces, brackets and parens means everything should be paired, other than merely {} [] and ().   So I listed a few commands expected to be auto-paired if possible. 

Even when braces, brackets and parens means only {}, [] and (), I would suggest that | | and \|\| are included, because they are kind of "mathematical parens". 

\verb is an exception when auto-pairing are not expected:  
- \verb(hello word), \verb[hello word], \verb{hello word}  *wouldn't* work, but
- \verb(hello word(, \verb[hello word[, \verb{hello word{  will.

r-stein:
@Romeoly For the quotes I have created the [LaTeXSmartQuotes](https://packagecontrol.io/packages/LaTeXSmartQuotes) package (this may be included into LaTeXTools some day). I don't think the \verb thing is useful, because you won't easily skip over the second delimiter, i.e. you type `\verb+test+` and get `\verb+test+|+` instead of `\verb+test+|`.

For the other commands I really ask how would you expect it to be handled, e.g. when you select the auto-complete entry, when you type the characters, or handle it in the env-close command (`C-l, .`) .

Romeoly:
@r-stein I agree with you that \verb is useless. So let's forget about it. 

Let ^ be the caret position, be cause I shall talk about |. 

I expect that 

| type | get |
| :-----:|:----:|
| \| | \| ^ \| |
|\\\||\\\| ^ \\\||
|\\left\||\\left\| ^ \\right\||
|\\left\\\||\\left\\\| ^ \\right\\\||
|\\vert|\\vert ^ \\vert|
|\\lvert|\\lvert ^ \\rvert|

And the backspace works similar as ( ^ ).

davidgontier:
@Romeoly I do not agree for "|" giving " | ^ |". A lot of people are using "|" for "such as", or for Dirac's notations.

Romeoly:
@davidgontier Imho, \mid may be a better choice for "such that". E.g. {x \mid x^2 < 1}. The spacing is better than {x | x^2 < 1 }. Speaking of Dirac's Bra and Ket notation, I am not sure about the best choice. It's a difficult choice.

tobiasdiez:
For the normal math inline environment (with the dollar symbols) you have the nice feature that if you are inbetween an empty pair (e.g, `$|$`) and press delete, then both dollar symbols are deleted. It would be nice if something similar could be added for the other paired brackets also.

Moreover, with the current code the input `\(` is converted to `\(|\)`, while in the above table you have `\( | \)` with additional spaces. I prefer the second version but understand that might be just a personal preference.

alexander-e:
I stumbled upon this pull request while looking for an easy way to get this behavior (i.e., the auto pairingn of `\(` and `\[`) myself. As far as I am concerned, this would be a major addition. Since it has been already implemented, I see no reason why this cannot be added in the following release. The auto pairing of \left? and \right? is not necessary for me. This functionality can be easily achieved in a LaTeX way by using  the paired delimiter functionality of the mathtools package.

r-stein:
@alexander-e Yes, you are right. It will be released with the next release. Further improvements can be done in an other PR.

Tanedo:
Is there advice for how to modify my keybindings to change the default behavior for curly braces in math mode? 

Given `$a$` , I'd like to be able to highlight the 'a' and press `{` to give `${a}$` rather than `$\{a\}$`. 

I couldn't figure how to to do this from peeking at the default LaTeXTools keybindings.

Tanedo:
My apologies! This was a conflict from a different package (`latexyz`) and has nothing to do with `LaTeXtools`.

For what it's worth: I added the following to my user keybindings to overwrite the unwanted `latexyz` behavior. 

```json
  { "keys": ["{"], "command": "insert_snippet", "args": {"contents": "{${0:$SELECTION}}"}, "context":
        [
             { "key": "selector", "operator": "equal", "operand": "text.tex.latex meta.environment.math" },
             { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
             { "key": "selection_empty", "operator": "equal", "operand": false, "match_all": true }
        ]
    },
```

randy3k:
Yap, it is the [keyblinding](https://github.com/randy3k/LaTeXYZ/blob/f324881f910946ffa2527fcf825d7e2a63a3131b/support/Default.sublime-keymap#L245-L251).

juanfal:
> My apologies! This was a conflict from a different package (`latexyz`) and has nothing to do with `LaTeXtools`.
> 
> For what it's worth: I added the following to my user keybindings to overwrite the unwanted `latexyz` behavior.
> 
> ```json
>   { "keys": ["{"], "command": "insert_snippet", "args": {"contents": "{${0:$SELECTION}}"}, "context":
>         [
>              { "key": "selector", "operator": "equal", "operand": "text.tex.latex meta.environment.math" },
>              { "key": "setting.auto_match_enabled", "operator": "equal", "operand": true },
>              { "key": "selection_empty", "operator": "equal", "operand": false, "match_all": true }
>         ]
>     },
> ```

Thank you! You saved me from a nervous breakdown!

PR #1204, opened 2017-08-25
Add command for toggling starred version of surrounding environment

vqv:
Builds on https://github.com/SublimeText/LaTeXTools/pull/585 by @r-stein to add a command for toggling the starred version of a surrounding environment with keybinding `C-l,C-*`.  This is useful for enabling/disabling numbering of math environments such as `equation` and `align`.  The implementation reuses existing logic in `latex_change_environment` for detecting environment regions so it should work with nested environments and multiple cursors.

Demo
toggle_starred_environments

### DIFF
--- a/Default (Linux).sublime-keymap
+++ b/Default (Linux).sublime-keymap
@@ -248,6 +248,11 @@ LaTeX Package keymap for Linux
 		"context":  [
 			{"key": "selector", "operator": "equal", "operand": "text.tex.latex"}],
 		"command": "latex_change_environment"},
+	// Toggle the starred version of environment
+	{ "keys": ["ctrl+l", "ctrl+shift+8"],
+		"context":  [
+			{"key": "selector", "operator": "equal", "operand": "text.tex.latex"}],
+		"command": "latex_toggle_environment_star"},
 
 	// Wrap selected text in emph, bold or underline
 	{ "keys": ["ctrl+l","ctrl+e"],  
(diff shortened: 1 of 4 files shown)


change_environment.py
```diff
@@ -6,78 +6,94 @@
...
         view.sel().clear()
         for r in new_regions:
             view.sel().add(r)
 
+class LatexToggleEnvironmentStarCommand(sublime_plugin.TextCommand):
+    def run(self, edit):
+        view = self.view
+        new_regions = _find_env_regions(view)
+        # 
+        if view.substr(new_regions[0])[-1] == '*':
```
r-stein: Can you please change this to `if view.substr(new_regions[0]).endswith('*'):` so it won't throw an error for empty strings. In addition you should add a safety check and return if `new_regions` is empty.

Default (Linux).sublime-keymap
```diff
@@ -248,6 +248,11 @@ LaTeX Package keymap for Linux
 		"context":  [
 			{"key": "selector", "operator": "equal", "operand": "text.tex.latex"}],
 		"command": "latex_change_environment"},
+	// Toggle the starred version of environment
+	{ "keys": ["ctrl+l", "ctrl+shift+8"],
```
r-stein: Not sure if we should use this keybinding, because it is specific to the US layout.

vqv: I'm not sure what a good alternative is.  I doesn't see like Sublime allows specifying `*` as a key.

r-stein: I also don't see a better alternative and it also works on other keyboards, so I think it is ok.

r-stein:
Thanks.

Issue #1199, opened 2017-08-20
Missing error: Incomplete \iftrue; all text was ignored after line ...


evandrocoan:
For the code:
```latex
\documentclass[10pt,a5paper,twoside]{memoir}

\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}

\usepackage[brazil]{babel}
\usepackage[showframe,pass]{geometry}

\newif\ifdebug
\debugtrue
% \debugfalse

\begin{document}

Arquivo \ifdebug 22:00 \else \currenttime \fi h do dia \ifdebug TODAY \else \today \if .

\end{document}
```

The log has the error:
```
)
! Incomplete \iftrue; all text was ignored after line 81.
<inserted text> 
                \fi 
<*> ./test2.tex
               
The file ended while I was skipping conditional text.
This kind of error happens when you say `\if...' and forget
the matching `\fi'. I've inserted a `\fi'; this might work.

! Emergency stop.
<*> ./test2.tex
               
*** (job aborted, no legal \end found)
```
Is not show, it shows nothing wrong, while no PDF is being created:
```
[Compiling D:\test2.tex]

Basic Builder: running pdflatex...done.

No errors. No warnings. No bad boxes.

[Done in 00:00:03 seconds!]
```

___

Related:

1. https://github.com/SublimeText/LaTeXTools/issues/1196 Missing statements on Undefined control sequence. [ }] log parsing

evandrocoan:
This seems to be fixed with https://github.com/SublimeText/LaTeXTools/pull/1208

Issue #1205, opened 2017-08-25
not cataching error for wrong syntax


daviehh:
minium example:
file `mwe.tex`:

```
\documentclass{revtex4-1}
\usepackage{ulem}

\begin{document}
\section\textit{\uwave{Introduction}}
\end{document}
```

running
`latexmk -cd -f -pdf -interaction=nonstopmode -synctex=1 mwe.tex` in terminal gives the error as expected:

```
! Incomplete \ifmmode; all text was ignored after line 5.
<inserted text> 
                \fi 
<*> mwe.tex
           
! Emergency stop.
<*> mwe.tex
```

but in latextools, this error is not caught, status bar says build complete with warning, and console in sublime says
```
Welcome to thread Thread-9
['latexmk', '-cd', '-f', '-pdf', '-interaction=nonstopmode', '-synctex=1', 'mwe.tex']
Running "latexmk -cd -f -pdf -interaction=nonstopmode -synctex=1 mwe.tex"
Finished normally
12
False True True
Expected PDF file /path/to/mwe.pdf not found
```

daviehh:
may be related to https://github.com/SublimeText/LaTeXTools/issues/1199

daviehh:
@r-stein `parseTeXlog.py` found the error, just not appearing in error summary. Here's the relevant part after running `python parseTeXlog.py /path/to/mwe.log`: (`parseTeXlog: Error found: ! Incomplete \ifmmode; all text was ignored after line 5.` but the `Errors:` section at the end is empty)

```
parseTeXlog: Error found: ! Incomplete \ifmmode; all text was ignored after line 5.
parseTeXlog: Reporting error in line: <inserted text> 
parseTeXlog: Reporting error in line:                 \fi 
parseTeXlog: Reporting error in line: <*> mwe.tex
parseTeXlog: Reporting error in line:            
parseTeXlog: Reporting error in line: The file ended while I was skipping conditional text.
parseTeXlog: Reporting error in line: This kind of error happens when you say `\if...' and forget
parseTeXlog: Reporting error in line: the matching `\fi'. I've inserted a `\fi'; this might work.
parseTeXlog: Reporting error in line: 
parseTeXlog: Reporting error in line: ! Emergency stop.
parseTeXlog: Emergency stop found
parseTeXlog: Reporting error in line: <*> mwe.tex
parseTeXlog: Reporting error in line:            
parseTeXlog: Reporting error in line: *** (job aborted, no legal \end found)
[...]

PR #1208, opened 2017-08-25
fix parsing for "Incomplete \if...; all text was ignored after line"

daviehh:
fix parsing for "Incomplete \if...; all text was ignored after line"
fix https://github.com/SublimeText/LaTeXTools/issues/1199 and https://github.com/SublimeText/LaTeXTools/issues/1205

### DIFF
--- a/parseTeXlog.py
+++ b/parseTeXlog.py
@@ -705,6 +705,16 @@ def parse_tex_log(data, root_dir):
 				errors.append(err_msg)
 				errors.append("Check the TeX log file for more information")
 				continue
+			#special: all text was ignored after line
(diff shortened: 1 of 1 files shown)


r-stein:
Thanks for your contribution. Since this does not seem to affect other parts of the code I will merge this for the next prerelease unless someone rejects it.

daviehh:
thanks. used the example in the original code starting
```
# Special error reporting for e.g. \footnote{text NO MATCHING PARENS & co
if "! File ended while scanning use of" in line:
```
since they have similar error structure.

r-stein:
Thanks.

PR #1211, opened 2017-09-06
change some code that makes latextools capable of dealing with markdo…

D-W-:
…wn files


D-W-:
preview md files

D-W-:
need to change `Markdown.sublime-syntax` and add block-math and inline-math in it to locate the math

PR #1224, opened 2017-09-16
The master has conflicts with version4 branch

evandrocoan:
My master is up to date with branch version4, which is not mergable with the current master branch https://github.com/evandrocoan/LaTeXTools/pull/2 after the update https://github.com/SublimeText/LaTeXTools/commit/b116fd56117fc5a1a48bce43f01c4eabf33fe27a, so my updates from this will be stuck until the merge of master version4 branch with the version4 branch happens.

The master branch should not have more updates then the other branches, as develop and version4. How can the stable version has problem fixes which are not presented on the development version? If something need to immediately be pushed to the stable version on the master branch, such fix should be also cherry picked for other branches as version4 and develop, so they are also fixed.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -8,7 +8,7 @@ by Ian Bacher, Marciano Siniscalchi, and Richard Stein
 **Documentation:**
 <https://latextools.readthedocs.io>
 
-*Latest revision:* v3.14.0 (2017-05-10).
+*Latest revision:* v3.14.1 (2017-06-01).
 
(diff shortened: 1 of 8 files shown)


ig0774:
Ok, but version4 is a live development branch. It's not intended for wide-spread use, nor necessarily to track all the changes in master (obviously things are different when someone is ready to spin-up a RC).

r-stein:
I agree with your point and was looking forward to merge the master into version4, but wanted to also include #1194 in the next prerelease. That prerelease is delayed, because I want to do some minor changes, but I will release it tomorrow with or without that PR.

PR #1228, opened 2017-09-19
Implemented the display of the last log lines on parseTeXlog.py as

evandrocoan:
they usually contains the errors as in:

LaTeX Font Info:    External font `cmex10' loaded for size
(Font)              <7> on input line 1.
LaTeX Font Info:    External font `cmex10' loaded for size
(Font)              <5> on input line 1.

test3.tex:23: Undefined control sequence

### DIFF
--- a/parseTeXlog.py
+++ b/parseTeXlog.py
@@ -5,6 +5,7 @@ import re
 import sys
 import os.path
 
+from collections import deque
 
 # To accommodate both Python 2 and 3
 if sys.version_info >= (3,):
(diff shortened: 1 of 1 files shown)


r-stein:
So you are appending the last 7 log lines if an fatal error is detected, but no errors are present. This sounds useful and I will merge this if nobody disagrees.

evandrocoan:
I think this is happening for me all the, after I set `"options" : ["-file-line-error", "-halt-on-error"],` so every time there are errors, they are on these last lines but not being showed by LatexTools.

ig0774:
I don't really see a problem with the PR, but `-file-line-error` in general will screw up the parsing.

PR #1229, opened 2017-09-20
Trimmed white spaces on logparse

evandrocoan:


r-stein:
Thanks for you effort, but I was looking forward to do such formatting changes shortly before the release. Otherwise they my conflict with other changes.

evandrocoan:
It certainly will conflict, so lets us wait until there before merging. If there are no conflicts we can just merge. If the conflicts are simple, we just fix them, otherwise discard this and perform the same changes again.

To do this changes fast I used a Sublime Text vanilla install, set the settings:
```javascript
{
	"ensure_newline_at_eof_on_save": true,
	"trim_trailing_white_space_on_save": true,
}
```
Then do a find all `Ctrl+Shif+F` with the:

1. Regex search `^\A(.*)$`
1. Folder `Packages\latextools`
1. Replace by `\1`

image

It will open all files from LatexTools with the buffer dirt, but not modification, then you just do the `File -> Save All` command. And Sublime Text will save all files trimming the whitespaces and ensuring a new line at the end of the file.

Issue #1235, opened 2017-09-25
TypeError: can only join an iterable


evandrocoan:
After changing the rulers on my project settings I got this on my console:
```
Traceback (most recent call last):
  File "D:\SublimeText\Data\Packages\LaTeXTools\st_preview\preview_math.py", line 415, in update_preamble_str
    self.preamble_str = "\n".join(self.preamble)
TypeError: can only join an iterable
reloading /D/User.sublime-project
```

The lines mentioned are:
```python
File: LaTeXTools/st_preview/preview_math.py
411:         def update_preamble_str(init=False):
412:             if isinstance(self.preamble, str):
413:                 self.preamble_str = self.preamble
414:             else:
415:                 self.preamble_str = "\n".join(self.preamble)
416: 
417:             if not init:
418:                 self.reset_phantoms()
```

___

Related:

1. https://github.com/Monnoroch/ColorHighlighter/issues/370 AttributeError: 'bool' object has no attribute 'find'

evandrocoan:
Now I after restarting Sublime Text I got the error on another line:
```python
File:  LaTeXTools/st_preview/preview_math.py
404:     def _init_watch_settings(self):
405:         # listen to setting changes to update the phantoms
406:         def update_packages_str(init=False):
407:             print( "self.packages: " + str( self.packages ) )
408:             self.packages_str = "\n".join(self.packages)
409:             if not init:
410:                 self.reset_phantoms()
```
But this time I added `print( "self.packages: " + str( self.packages ) )` before it, and this was the result:
```
self.packages: 0
Traceback (most recent call last):
  File "D:\User\Dropbox\Applications\SoftwareVersioning\SublimeText\Data\Packages\LaTeXTools\st_preview\preview_math.py", line 408, in update_packages_str
    self.packages_str = "\n".join(self.packages)
TypeError: can only join an iterable
```

r-stein:
What is the content of your [preview_math_template_preamble](https://github.com/SublimeText/LaTeXTools/blob/st3-3.14.2/LaTeXTools.sublime-settings#L143) setting (and the packages for the other issue)?

evandrocoan:
It is:
```javascript
	"preview_math_template_preamble": "",
```
 But what is his packages settings? Is this:
```javascript
	"preview_math_template_packages": [
		"\\usepackage{amsmath}",
		"\\usepackage{amssymb}",
[...]

PR #1236, opened 2017-09-26
Typeerror can only join an iterable

evandrocoan:
Fix https://github.com/SublimeText/LaTeXTools/issues/1235 TypeError: can only join an iterable

### DIFF
--- a/st_preview/preview_math.py
+++ b/st_preview/preview_math.py
@@ -414,7 +414,7 @@ class MathPreviewPhantomListener(sublime_plugin.ViewEventListener,
             if isinstance(self.preamble, str):
                 self.preamble_str = self.preamble
             else:
-                self.preamble_str = hasattr(self.preamble, "__iter__") if "\n".join(self.preamble) else ""
+                self.preamble_str = "\n".join(self.preamble) if hasattr(self.preamble, "__iter__") else ""
 
             if not init:
                 self.reset_phantoms()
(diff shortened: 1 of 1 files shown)


r-stein:
I am not sure whether we should start to guard against every misconfiguration.

evandrocoan:
What is the misconfiguration which causes this?

r-stein:
This just reads and joins the values of [`preview_math_template_packages`](https://github.com/SublimeText/LaTeXTools/blob/st3-4.0.0-alpha.5/LaTeXTools.sublime-settings#L142), which must be an array (as stated in the comment above it) and `preview_math_template_preamble`, which can be an array or an string (as stated in the comment above it).
Also the math scope is set as a string in the setting [`preview_math_scope`](https://github.com/SublimeText/LaTeXTools/blob/st3-4.0.0-alpha.5/LaTeXTools.sublime-settings#L117) and also an empty string should work (but is not useful) to preview the whole document.

evandrocoan:
So the error is because the settings are empty strings?

I have not changed these default settings. Is their default value a invalid value for them?

If these lines are just reading the these settings, the settings should be validated while they are being read, then I should patch elsewhere.

ig0774:
The links in [@rstein's comment above](https://github.com/SublimeText/LaTeXTools/pull/1236#issuecomment-334787427) point to the default values, neither of which is an empty string. The only way for the value to have changed is if you edited the local settings, which would give the behaviour you're seeing.

evandrocoan:
When I said about empty string I was talking about the value of `preview_math_template_preamble` which @r-stein mentions on its comment.

r-stein:
If you open the ST console and write 
```py
from LaTeXTools.latextools_utils.settings import get_setting; get_setting("preview_math_template_packages")
```
what output do you get?

PR #1263, opened 2017-11-27
Fix bug where unfinished lines were not being detected

charlesbaynham:
The unfinished line detection code was never triggering, causing `(Log parsing issues. Disregard unless something else is wrong.)` errors. 

For an MWE, try compiling this on Windows:

```
\documentclass{article}

\usepackage{pgfcore}

\begin{document}

If run on Windows, the log file will now contain several lines (e.g. l.150, l.160) that have only a single quote mark. Try searching for 
\verb¬^[^\n"]*"[^\n"]*$¬ to find them. 

\end{document}
```


r-stein:
Looks good to me I will merge it before the next release.

PR #1267, opened 2017-12-11
Strip quotes from filename when analyzing for cache

sethblumberg:
This commit strips any surrounding doublequotes from each filename that is added to the cache.

Files with spaces in the filename need surrounding quotes for \input{} to work.  With this change, things like labels inside them will show up in the reference autocomplete pop-up list.

This change isn't comprehensive, so feel free to write a better one.  For example, `C+l,C_o` (to open a file if it exists, or make a new one if it doesn't exist) still trips up on quotes surrounding a  filename.


r-stein:
Sounds reasonable. I will have a look at it in the next days.

sethblumberg:
I made a similar change in `jumpto_tex_file.py` over in pull request [1270](https://github.com/SublimeText/LaTeXTools/pull/1270).  They're separate because I didn't know how to add another change to this pull request.

PR #1270, opened 2017-12-15
strip doublequotes from filename in _jumpto_tex_file

sethblumberg:
Allows `\input{"../my folder/my file.tex"}` to be read as `\input{../my folder/my file.tex}`.  The former is necessary to have filenames with spaces to be typeset by LaTeX, but without this change, the command _jumpto_tex_file ends up creating a folder called `"..` (that's doublequote then dot dot) in the home folder.


sethblumberg:
This change is similar to the one in my pull request [1267](https://github.com/SublimeText/LaTeXTools/pull/1267), which striped quotes from filenames when analyzing files for the cache.  They're separate pull requests because I didn't know how to add this change to the existing pull request.

PR #1300, opened 2018-03-19
Fix for 1298

### LINKED ISSUE #1298: Custom builder not running
Hi, after an update (a few weeks ago), I have issues with biber. I am running Sublime Text 3.0 on Ubuntu 16.04, and using lualatex + biber. I suspect I have the same issues as outlined here:
https://github.com/SublimeText/LaTeXTools/issues/637
https://tex.stackexchange.com/questions/286095/biber-and-sublime-text

As suggested in the above threads, I want to run a custom builder (biberBuilder, see https://github.com/SublimeText/LaTeXTools/issues/637#issuecomment-192361009). `Preferences|Browse Packages` tells me that my packages folder is at `~/.config/sublime-text-3/Packages`, so I have included` biberBuilder.py` in `~/.config/sublime-text-3/Packages/User/LaTeXTools/builders/`. When I run `window.open_file(sublime.packages_path() + "/User/LaTeXTools/builders/biberBuilder.py")` in the console the correct file `biberBuilder.py` is opened, so it seems to be in the right place.
Moreover, I have added the following to my `.sublime-project` file:
```
"builder_settings": {
            "builder": "biber",
            "builder_path": "User/LaTeXTools/builders",
            "program": "lualatex",
            "options": ["--shell-escape"]
        }
```
However, when I do Ctrl+B, the basic builder runs, not my custom biber builder. 
What could be the issue? How do I troubleshoot this problem?

ig0774:
Makes the regex for the basic builder more permissive. Should fix #1298.

### DIFF
--- a/builders/basicBuilder.py
+++ b/builders/basicBuilder.py
@@ -33,7 +33,7 @@ from latextools_utils.external_command import external_command, get_texpath
 
 # Standard LaTeX warning
 CITATIONS_REGEX = re.compile(
-    r"Warning: Citation [`|'].+' (on page \d+)? undefined")
+    r"Warning: Citation [`|'].+' (?:on page \d+ )?undefined")
 # Capture which program to run for BibLaTeX
 BIBLATEX_REGEX = re.compile(
     r"Package biblatex Warning: Please \(re\)run (\S*)")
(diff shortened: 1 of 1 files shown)


dafrick:
I think your fix is almost correct, it should be
`r"Warning: Citation [`|'].+' (on page \d+ )?undefined")`
instead of
`r"Warning: Citation [`|'].+' (on page \d+)? undefined")`
(otherwise you have one whitespace too many)

ig0774:
@dafrick: Good catch!

PR #1324, opened 2018-05-10
Add sec, fig, app, tab, alg to triggers for ref

### LINKED ISSUE #1323: Feature request: extend \ref to custom commands like \figref or \secref
When writing a huge document it is really helpful to define the way figures, sections, etc. are referenced in a single place to allow changing this at any point in time. Currently, LaTeXTools works with `\ref` and `\eqref`, but won't work with custom commands like `\secref` for sections or `\figref` for figures. Is it possible to extend the completion functionality to these too?

UPD: I have added a small PR to solve this issue. It would probably make more sense to add a setting for that, but I don't have the time now to look into how to do it here. I might return to this later if somebody finds it a useful feature.

niosus:
We use the following snippet in out Latex documents: 
```latex
\newcommand{\secref}[1]{Section~\ref{#1}}
\newcommand{\figref}[1]{Figure~\ref{#1}}
\newcommand{\tabref}[1]{Table~\ref{#1}}
\newcommand{\appref}[1]{Appendix~\ref{#1}}
\renewcommand{\eqref}[1]{Equation~(\ref{#1})}
\renewcommand{\algref}[1]{Algorithm~\ref{#1}}
```

This PR extends the list of valid prefixes before `\ref` to the ones mentioned above. Close #1323

### DIFF
--- a/latex_ref_completions.py
+++ b/latex_ref_completions.py
@@ -19,7 +19,7 @@ _ref_special_commands = "|".join([
     "", "eq", "page", "v", "V", "auto", "autopage", "name",
     "c", "C", "cpage", "Cpage", "namec", "nameC", "lcnamec", "labelc",
     "labelcpage", "sub", "f", "F", "vpage", "t", "p", "A", "B", "P", "S",
-    "title", "headname", "tocname"
+    "title", "headname", "tocname", "sec", "fig", "app", "tab", "alg"
 ])[::-1]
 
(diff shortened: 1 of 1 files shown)


niosus:
Closing in favor of the solution discussed in the issue here: https://github.com/SublimeText/LaTeXTools/issues/1323#issuecomment-388161392

Issue #1329, opened 2018-05-28
NeoVintageous messes with Quickpanel input{} completion


DrBones:
When typing `\input{` and then choosing something from the quickpanel that pops up I get a correct completion but at one character earlier when also using NeoVintageous, e.g  `\inpu{CompletedFilenameHere}t`
This also happens with all other command that use the quickpanel, like `\cite{` or `\ref{`

This might be more of a bug in Neovintageous but it only occurs in combination with LatexTools.  I believe this is rather recent as I use the autocomplete extensively.

Also, when no other extensions or custom config is used.

System:
Windows 10
Sublime Build 3176
LatexTools 3.15.1
NeoVintageous 1.6.2

r-stein:
That's weird can you open the ST console and log commands (`sublime.log_commands(True)`) and inputs (`sublime.log_input(True)`) when this happens?

DrBones:
Hi, that was fast. Sure! 

The console throws a lot of ghostscript and external_command.py errors.

Here is the full output

```
chr evt: i (0x69)
chr evt: n (0x6e)
chr evt: p (0x70)
chr evt: u (0x75)
chr evt: t (0x74)
Error finding Ghostscript version for None
Traceback (most recent call last):
  File "C:\Users\username\Downloads\Sublime Text Build 3176 x64\Data\Packages\LaTeXTools\st_preview\preview_utils.py", line 99, in _update_gs_version
    raw_version = check_output([_GS_COMMAND, '-version'])
  File "C:\Users\username\Downloads\Sublime Text Build 3176 x64\Data\Packages\LaTeXTools\latextools_utils\external_command.py", line 348, in check_output
    show_window=show_window
  File "C:\Users\username\Downloads\Sublime Text Build 3176 x64\Data\Packages\LaTeXTools\latextools_utils\external_command.py", line 268, in execute_command
    show_window=show_window
  File "C:\Users\username\Downloads\Sublime Text Build 3176 x64\Data\Packages\LaTeXTools\latextools_utils\external_command.py", line 181, in external_command
    if not os.path.isabs(command[0]):
  File "./python3.3/ntpath.py", line 102, in isabs
  File "./python3.3/ntpath.py", line 162, in splitdrive
TypeError: object of type 'NoneType' has no len()
key evt: control+alt+7
chr evt: { (0x7b)
command: latex_fill_all {"completion_type": "input", "insert_char": "{"}
Error finding Ghostscript version for None
[...]

PR #1355, opened 2018-07-31
Workaround for #1329.

SolidTux:
Workaround for inserting issue when combined with NeoVintageous.


SolidTux:
This still has problems as discussed on the issue page and therefore probably is not ready for merging.

PR #1395, opened 2019-02-15
Add stdin=subprocess.PIPE to prevent fail on win

apetri:
I modified makePDF.py adding a stdin=subprocess.PIPE argument to external_command(); without this building fails on Windows 7

### DIFF
--- a/makePDF.py
+++ b/makePDF.py
@@ -311,6 +311,7 @@ class CmdThread ( threading.Thread ):
 							cmd,
 							env=env,
 							use_texpath=False,
+							stdin=subprocess.PIPE,
 							stdout=subprocess.PIPE,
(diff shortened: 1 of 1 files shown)


r-stein:
Not sure if this leads to new problems. This build version worked for me on windows 7.

apetri:
For some reason it still doesn’t work on mine, subprocess.Popen fails without the stdin argument. It could be my win 7 build

deathaxe:
Resolved conflicts and merged as https://github.com/SublimeText/LaTeXTools/commit/2e3c39ef319f9b771ae387b5395ab7cf801127c6.

Most likely no longer relevant, but back in the days this patch needed to be applied to various plugins to fix occational issues with starting external processes.

PR #1401, opened 2019-04-05
Add root folder to TEXINPUTS for preview-math

bordaigorl:
Currently the Math Preview feature does not support including external files.
This is because:

1. the preview is built in a separate temp folder and these files are not copied there
2. changes to these imported files would not trigger updates to the preview

Assuming 2 can be solved by a disclaimer in the documentation (the use case is to import custom packages or custom macros which are seldom updated), this patch solves problem 1 by augmenting the environment variable `TEXINPUTS` with the folder of the current root of the document.

### DIFF
--- a/st_preview/preview_math.py
+++ b/st_preview/preview_math.py
@@ -12,6 +12,8 @@ import types
 import sublime
 import sublime_plugin
 
+from .. import getTeXRoot
+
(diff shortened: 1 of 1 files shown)


bordaigorl:
Closing because isomorphic to #1179 😭

PR #1421, opened 2019-08-05
Fixed a reference to set_async_timeout (should be set_timeout_async).

anthonyjclark:
A reference to `set_async_timeout` should be `set_timeout_async`. I only found this typo in `latextools_utils/sublime_utils.py`. For more information see these:

- https://github.com/SublimeTextIssues/Core/issues/825
- https://www.sublimetext.com/docs/3/api_reference.html

All other uses of the timeout are correct as far as I can tell.


r-stein:
You are correct, good catch.

anthonyjclark:
Thanks for maintaining such a great tool.

PR #1454, opened 2019-12-05
blackslash is now handled correctly in sublime text 4

randy3k:
as title


randy3k:
Thanks. By the way, what is the status of LaTeXTools v4?

r-stein:
Unfortunately it is very delayed, because of a lack of time/motivation. My current goal is to do the incompatible changes (as in #1423) and then release v4 in the current feature state and add new features afterwards.

Issue #51, opened 2012-07-13
View PDF and Jump to PDF fail to fail gracefully on Windows


badge:
I've been rooting around in the LaTeXTools package (wanting to implement TeXCount.pl), and discovered that, on Windows, when a non-.tex file is in the open view, the View PDF and Jump to PDF commands fail to warn the user as expected. Instead, the following is printed to the console:

```
Traceback (most recent call last):
File ".\sublime_plugin.py", line 339, in run_
File ".\viewPDF.py", line 15, in run
AttributeError: 'View' object has no attribute 'fileName'
```

Sadly my lack of knowledge of Windows (usually on *nix!) and Python (usually in Perl!) make me unable to work out what's going on.

PR #1472, opened 2020-04-24
Fix error message when jumping to pdf from non-.tex-file

rensbaardman:
There was an ST1-syntax being used (the same as in https://github.com/SublimeText/LaTeXTools/issues/51) that caused a Python error. This PR fixes that, so when you are in a non-`.tex`-file and try to jump to the PDF, it gives a Sublime error.

This still leaves another question for me: I encountered this bug when I tried to recompile from a non-`.tex`-file. In that case, I think there should not be an error message, but instead it should either:
- reload the pdf on the same place if it was already opened
- open the pdf on the first page if it hadn't been open
Should I open an issue for this?

### DIFF
--- a/jumpToPDF.py
+++ b/jumpToPDF.py
@@ -117,7 +117,7 @@ class JumpToPdf(sublime_plugin.TextCommand):
 		if not is_tex_file(view.file_name()):
 			sublime.error_message(
 				"%s is not a TeX source file: cannot jump." %
-				(os.path.basename(view.fileName()),))
+				(os.path.basename(view.file_name()),))
 			return
 
 		root = getTeXRoot.get_tex_root(view)


r-stein:
Thanks

rensbaardman:
@r-stein I actually did try to fix the open question I posed: see the new commits.

It solved my problem and I think this should work in all situations, but I'm not familiar enough with the code base to be completely sure. Could you review it?

ig0774:
@rensbaardman I think I might be willing to merge this, but it's awfully hard to follow what was actually changed in this PR. Could you point out the relevant lines if you remember?

rensbaardman:
@ig0774 My apologies, looking back this is certainly not a clean PR, and I should have split up the cleaning and the fix. The fix itself can be found [in these lines](https://github.com/SublimeText/LaTeXTools/pull/1472/commits/7d454cb5f3273f56d9acf19e2c84caa1fda0a7d8#diff-eea05632775bd5ff7c1498c320c786a931edc0b06086c64c33fba37cf1a647feR115-R125).

PR #1495, opened 2020-09-15
Increased support for MiKTeX

maboloshi:
- MiKTeX has been supported on Windows, Linux, and OSX since `2.9.6300`. Therefore, information about its usage has been added.
- Linux and OSX install the Perl interpreter by default, so only if MiKTeX/Windows loses the Perl interpreter will the `traditional` builder use compiler driver `texify`.


ig0774:
@maboloshi Sorry this is quite a late review 😬. One thing I'm unsure of in this: does MiKTeX now install latexmk by default? The real reason we were using texify with MiKTeX is that is was installed and available by default and latexmk wasn't (I never had any issues using latexmk with MiKTeX on Windows... there was just the extra install step which also installed the Perl interpreter, etc.).

maboloshi:
@ig0774 In fact, `latexmk` and the perl interpreter need to be installed additionally, as mentioned in the `docs`.

deathaxe:
If pre-conditions are met (latexmk and perl are present) latexmk can be used with MikTeX by just setting

```
	"builder_settings": {
		"command": "latexmk",
	}
```

There's probably no need to change the default builder `texify`, automatically.

PR #1507, opened 2020-12-11
Added a Google Chrome viewer with forward sync

zsiegel92:
I added a viewer in `viewers/chrome.py` that mimics the structure of the `okular` and `evince` viewers. I tried not to use too many weird methods, though I suppose the applescript I wrote may require user permissions. It works for me!


zsiegel92:
I think the Chrome viewer I wrote is the only viewer that directly calls `synctex` (rather than delegating that task to Skim or Zathura, etc.). It calls `/Library/TeX/texbin/synctex`.

If anyone knows the best practice for finding the synctex binary on a given machine, I'll replace it with that and I think this will be pretty merge-able!

deathaxe:
As Chrome is supported on all platforms, a viewer plugin should also work on Linux/Mac/Windows.

PR #1533, opened 2021-08-23
Fix critical mistake in parseTeXlog.py

Jollywatt:
For parsing over/underfull messages, there is a line (504) that tests `line == " []"` — which is clearly meant to read `line[-3:] == " []"`, since `line` is a string containing the _entire_ line. (There are four other occurrences of `line[-X:] == "…"` in this file.)

Please accept this PR! It was impossible for me to continue working with LaTeXTools when log parsing was broken, but luckily the fix was trivial! :)

### DIFF
--- a/parseTeXlog.py
+++ b/parseTeXlog.py
@@ -501,7 +501,7 @@ def parse_tex_log(data, root_dir):
 				# Sometimes it's " []" and sometimes it's "[]"...
 #				if len(line)>0 and line[:3] == " []" or line[:2] == "[]":
 				# NO, it really should be just " []"
-				if len(line)>0 and line == " []":
+				if len(line)>0 and line[-3:] == " []":
 					ou_processing = False
 				else:
 					current_badbox += line


parseTeXlog.py
```diff
@@ -501,7 +501,7 @@ def handle_badbox(l):
 				# Sometimes it's " []" and sometimes it's "[]"...
 #				if len(line)>0 and line[:3] == " []" or line[:2] == "[]":
 				# NO, it really should be just " []"
-				if len(line)>0 and line == " []":
+				if len(line)>0 and line[-3:] == " []":
```
keith-hall, line 504: have you considered using `endswith` instead of slicing, in case the line is shorter than 3 chars (not sure if such a case could happen but imo its more readable)? I'm not a user of this plugin, but can merge if someone chimes in to say it works...

ig0774: So, the log parsing is my least favourite part of the code, in part because LaTeX’s log output is highly sensitive to the set of packages and versions one uses so, e.g., this line probably works for your document setup but the existing version probably works just fine for some users and likely was designed to work perfectly with prior versions of LaTeX or particular packages.

I bring this up for two reasons. First, we definitely need some testing to be sure this doesn’t break things for other users. Additionally, it would be better if this patch was in the form:

```suggestion
				if len(line)>0 and (line == " []" or line.endswith(“ []”):
```

As that is less likely to break things for others.

Jollywatt: @keith-hall Indeed I did consider `.endswith`, but the `[-X:]` (anti?)pattern appears throughout the code, so I tried to blend in!
(Even for lines of length `< 3`, the two methods behave identically (assuming `type(line) == str`) — but yes, `.endswith` is more readable. Should I propose to replace all `line[-X:] == "…"` / `line[:X] == "…"` occurrences with `line.endswith` / `line.startswith`?)

---

@ig0774 I agree, this should be handled with extreme care… is there a system in place for testing `parseTeXlog.py` against sample logs? 

In the case at hand, your suggested change `line == " []" or line.endswith(" []")` is exactly equivalent to `line[-3:] == " []"` (assuming `type(line) == str`), so that’s ok.

---

The only possible danger in this PR is if relaxing the condition `line == " []"` to `line.endswith(" []")` causes “false positives”. This is possible, because TeX’s “over/underfull” sections can have multiple lines of arbitrary text — so the probability of a line happening to end with ` []` is nonzero. Luckily, if such a false positive occurred, then all that would happen would be that the `current_badbox` message gets truncated at bit early (and, at the moment, I don’t believe these ever get shown to the user anyway).

If we do not relax the condition `line == " []"` to `line.endswith(" []")`, then, in strange cases like mine where `" []"` does not occur on its own line, the `current_badbox` gobbles up the remaining log file and errors ungracefully.

Neither is perfect, but I believe this PR is the better option.

---

I realise my log is probably the unusual one here. My `.log` typically looks like this:

```
Overfull \hbox (68.93678pt too wide) has occurred while \output is active
\TU/LibertinusSerif(0)/m/it/10 SURJECTIVITY AND NON-INJECTIVITY OF THE EXPONENT
IAL MAP IN SPACETIME \TU/LibertinusSerif(0)/m/n/10 29

\hbox(6.77+1.99998)x280.0, glue set - 1.0 []

[29] [30] [31]
Overfull \hbox (1.79085pt too wide) in paragraph at lines 1521--1523
[]\TU/LibertinusSerif(0)/m/n/10 In the case of Lorentz transformations $[][][][
][] [][]$, some closed-form

\hbox(9.03581+1.86)x280.0, glue set - 1.0 []
```

…you’ll notice the over/underfull sections are terminated with a line **ending** with ` []`.
Whereas, [this example log](https://gist.github.com/Capelare/1d621611c40178099471#file-gistfile1-txt-L1539) found on issue #104 looks like this.

```
Overfull \hbox (6.50519pt too wide) in paragraph at lines 24--25
[]\OT1/cmr/m/n/12 Los pa-que-tes en ROS se crean f^^Sacil-men-te con he-rra-mie
n-tas co-mo \OT1/pcr/m/n/12 roscreate-pkg\OT1/cmr/m/n/12 ,
 []


Overfull \hbox (9.07646pt too wide) in paragraph at lines 33--34
[]\OT1/pcr/m/n/12 CMakeLists.txt\OT1/cmr/m/n/12 : Ar-chi-vo de com-pi-la-ci^^So
n CMa-ke (\OT1/cmr/m/it/12 Cross Plat-form Ma-ke \OT1/cmr/m/n/12 [  []  36[]]).
 
 []
```

Maybe this log, where the ending line is **exactly** ` []`, is more typical… not sure.

PR #1592, opened 2024-01-24
Added support for Sioyek viewer

guigui245:
Support for Sioyek viewer in windows from https://github.com/ahrm/sioyek  .
As far as I'm aware this is the only viewer in windows capable of overviews hyperlinks inside a pdf documents like evince does in linux.


scillidan:
Hope add it

PR #1598, opened 2024-11-16
grammar

austinwmille:
deleted a single character to fix plurals. i did not hate it the first way, so maybe intentional?


austinwmille:
This is literally my first GitHub edit, and now I don't even want to do it. I just was reading a doc and paused (over a single letter 's') and clicked a button and now it wants to send people messages. Feel free to ignore this

PR #1640, opened 2025-08-24
Bug fix: return aux dir in get_aux_directory for traditional builder

alberti42:
# Description

This PR fixes a bug in the `get_aux_directory` function in `./latextools/utils/output_directory.py` where the **traditional builder** (based on TeXLive’s `latexmk` on macOS/Linux or MiKTeX’s `texify` on Windows) was not properly handled.

Both TeXLive and MiKTeX support the `--aux-directory` flag, but the function was incorrectly returning `None` for traditional builds. This change ensures that the auxiliary directory is returned when using the traditional builder, if the user has set it.

---

# Changes Made

* Added explicit detection of the **traditional builder**:
    
    ```python
    using_traditional = (get_setting("builder", "traditional") == "traditional")
    ```
* Updated condition in `get_aux_directory` to include `using_traditional` so that the auxiliary directory is returned in supported cases.

---

# Updated Function

```python
def get_aux_directory(view_or_root, return_setting=False):
    # traditional builder is based on TeXLive latexmk (macOS/Linux) or MiKTeX texify (Windows) — both distros (builders) support `--aux-directory` flag
    using_traditional = (get_setting("builder", "traditional") == "traditional")  # detect the builder

    # not supported using texify or the simple builder
    if not (using_miktex() or using_traditional) or using_texify_or_simple():
        if return_setting:
            return (None, None)
        else:
            return None
```

---

# Rationale

* The bug prevented the proper use of `--aux-directory` with the **traditional builder**, even though it is supported by the underlying tools (`latexmk` and `texify`).
* This fix improves consistency with the documentation and ensures users can leverage auxiliary directories as expected when using traditional builds.

---

# Testing

* Verified on:
    * **TeXLive (macOS/Linux)** with `latexmk`
* Confirmed that auxiliary directory paths are correctly returned when set by the user.

### DIFF
--- a/latextools/utils/output_directory.py
+++ b/latextools/utils/output_directory.py
@@ -29,8 +29,11 @@ class UnsavedFileException(Exception):
 # return_setting indicates that the raw setting should be returned
 # as well as the auxiliary directory
 def get_aux_directory(view_or_root, return_setting=False):
+    # traditional builder is based on TeXLive latexmk (macOS/Linux) or MiKTeX texify (Windows) — both distros (builders) support `--aux-directory` flag
+    using_traditional = (get_setting("builder", "traditional") == "traditional")
+
     # not supported using texify or the simple builder
-    if not using_miktex() or using_texify_or_simple():
+    if not (using_miktex() or using_traditional) or using_texify_or_simple():
         if return_setting:
             return (None, None)
         else:


deathaxe:
TBH, the whole thing seems weird and I don't think `get_aux_directory()` and `get_output_directory()` should attempt to make assumptions on installed distro and used builder plugin at all. Their function is to provide directory names from directives or settings.

The calling builder plugin has to check and handle supported features of underlying distro/builder and decide how to make use of aux/output directory.

Hence instead of possibly further increasing confusions, I prefer to refactor builders in ways to always put temporarily files into aux and possibly copy final document over to specified output directory.

- latexmk, used by traditional builder, supports --aux-directory arugment.
- basic builder directly runs pdflatex, directly. It can be tought to support separate aux and output directories as well
- script builder provides variables for both, but they currently don't seem to be passed in due to those distro checks in get_aux_directory()

FWIW, TeXLive also works fine on Windows.

alberti42:
I agree that the current implementation (regardless of my patch) is not the cleanest. I believe the reason is that who originally implemented it decided not to provide support for auxiliary directories for pdflatex or lualatex because these flags are not directly supported by these two commands. However, I agree with you, in principle, one can work out a clean implementation that adds support for the auxiliary directory to pdflatex, lualatex and similar builders as well. But this needs to come on top of the existing commands—nothing that Python and Sublime could not handle.

My patch was less ambitious. I wanted to restore support for the auxiliary directory for latexmk (traditional builder) since this flag is well documented and supported by the distro.

> FWIW, TeXLive also works fine on Windows.

Fantastic. I did not know it. The last Windows version I worked with was Windows 95 ;-)

Do you think you can accept my patch, or do you prefer to first code a clean implementation supporting the auxiliary directory for all builders?

deathaxe:
I am fine to merge the PR to honor your efforts, but #1641 will effectively drop your changes as I was keen enough to start cleaning up code while thinking about your proposal. Primarily as I got confused by the miktex vs. builder logic, which I find hard to follow.

PR #1642, opened 2025-08-28
Fix SumatraPDF not reloading re-build PDFs

### LINKED ISSUE #1373: Sumatra PDF does not reload document on compile
The first time I compile a LaTeX document, the PDF opens in Sumatra PDF, scrolled to the position of my furthest edit since my last compilation. After this, if I make changes to my LaTeX document and compile it again, the section of the PDF where my edit was performed is scrolled into view and highlighted momentarily in Sumatra PDF, but the new PDF isn't actually loaded. I have to manually reopen the document in Sumatra PDF in order to see my changes.

Sublime Text 3.1.1, Build 3176
Sumatra PDF 3.1.2
LaTeXTools 3.15.1
Windows 10 Enterprise (1709)

deathaxe:
fixes #1373

PR #1557 refactored "basic" builder to properly support "aux_directory" and "output_directory", which results in SumatraPDF correctly reloading any re-build document as expected.

latexmk supports `--aux-directory` argument to specify separate directory for temporary or intermediate build assets. The strategy it uses to move final documents to `--output-directory` however breaks file notifications and file reloading of SumatraPDF and possibly even fails with other viewers, which hold a lock on open PDF documents.

Hence this commit applies same strategy used for "basic" builder to "traditional" builder to workaround this restriction. The way how python's `shutil.move()` handles file transfer, in case target already exists, seems to be a better fit for Windows OS.

### DIFF
--- a/plugins/builder/basic_builder.py
+++ b/plugins/builder/basic_builder.py
@@ -140,10 +140,13 @@ class BasicBuilder(PdfBuilder):
         if self.aux_directory_full and self.aux_directory_full != dest_dir:
             for ext in (".synctex.gz", ".pdf"):
                 name = self.base_name + ext
-                shutil.move(
-                    os.path.join(self.aux_directory_full, name),
-                    os.path.join(dest_dir, name)
-                )
+                try:
+                    shutil.move(
+                        os.path.join(self.aux_directory_full, name),
+                        os.path.join(dest_dir, name)
+                    )
+                except OSError:
+                    pass
 
     def log_output(self):
         if self.display_log:
--- a/plugins/builder/traditional_builder.py
+++ b/plugins/builder/traditional_builder.py
@@ -1,4 +1,6 @@
+import os
 import shlex
+import shutil
 import sublime
 
 from .pdf_builder import PdfBuilder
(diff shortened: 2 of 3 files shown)


plugins/builder/traditional_builder.py
```diff
@@ -97,28 +99,26 @@ def commands(self):
...
-                    cmd.append("--aux-directory=" + self.aux_directory)
-
-                if self.output_directory is not None:
-                    # DO NOT ADD QUOTES HERE
-                    cmd.append("--output-directory=" + self.output_directory)
+                if self.aux_directory:
+                    # Don't use --aux-directory as the way how latexmk moves
+                    # final documents to possibly defined --output-directory
+                    # prevent file-reloading in SumatraPDF or even fail at all,
+                    # if document is opened and locked by viewer on Windows.
```
michaelblyons, line 108: I don't understand the comment from 105 to 108.

Is this close to what you mean?

```
                    # Don't use --aux-directory as the way latexmk moves
                    # final documents to a possibly defined --output-directory
                    # prevents files reloading in SumatraPDF or even fails
                    # if documents are opened and locked by viewer on Windows.
```

plugins/builder/basic_builder.py
```diff
@@ -140,10 +140,13 @@ def commands(self):
...
-                shutil.move(
-                    os.path.join(self.aux_directory_full, name),
-                    os.path.join(dest_dir, name)
-                )
+                try:
+                    shutil.move(
+                        os.path.join(self.aux_directory_full, name),
+                        os.path.join(dest_dir, name)
+                    )
+                except OSError:
```
michaelblyons, line 148: Can you add a comment why the move is able to fail.silently?

michaelblyons (commented):
I'm going to start using your `*args` pattern a couple places where I had been using lists. 🙂

plugins/builder/traditional_builder.py
deathaxe: I should've known, you are watching :-D

Yes it is.

PR #1657, opened 2025-09-21
Improve consistency and usability of build panel handling

alberti42:
# Description

This PR makes three small but related improvements around the LaTeXTools build panel:

* **Expose "Show build panel" in the Command Palette**
    * Adds a new command palette entry:
        
        ```
        makefileCopy codeLaTeXTools: Show build panel
        ```
    * This provides a discoverable way for users to reopen the build panel manually (useful when `"hide_build_panel"` is set to `"always"` or `"never"`).
* **Clarify and correct documentation of `hide_build_panel`**
    * The previous comments in `LaTeXTools.sublime-settings` were inconsistent:
        * The code fallback was `"no_warnings"`.
        * The defaults file set `"no_badboxes"`.
        * The comments claimed `"never"` was the default.
    * The docs are now clearer, consistent, and explicitly state the true default.
* **Align code fallback with shipped defaults**
    * In `make_pdf.py`, the fallback for `get_setting("hide_build_panel", ...)` is updated from `"no_warnings"` → `"no_badboxes"`.
    * This ensures consistency between the source code and the default settings file.

---

# Benefits

* Users get an easy, discoverable way to open the build panel on demand.
* Documentation is clearer and avoids contradictory information about defaults.
* Settings behavior is more consistent between code and defaults, reducing confusion for both users and contributors.

### DIFF
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -86,6 +86,11 @@
 		"args": { "hide_annotations_only": true }
 	},
 	{
+    "caption": "LaTeXTools: Show build panel",
+    "command": "show_panel",
+    "args": { "panel": "output.latextools" }
+	},
+	{
 		"caption": "Preferences: LaTeXTools Settings",
 		"command": "edit_settings",
 		"args": {
--- a/Default.sublime-commands
+++ b/Default.sublime-commands
@@ -86,9 +86,9 @@
 		"args": { "hide_annotations_only": true }
 	},
 	{
-    "caption": "LaTeXTools: Show build panel",
-    "command": "show_panel",
-    "args": { "panel": "output.latextools" }
+		"caption": "LaTeXTools: Show build panel",
+		"command": "show_panel",
+		"args": { "panel": "output.latextools" }
 	},
 	{
 		"caption": "Preferences: LaTeXTools Settings",
--- a/LaTeXTools.sublime-settings
+++ b/LaTeXTools.sublime-settings
@@ -465,18 +465,18 @@
 	// Scroll build output panel to top after build finished.
 	"scroll_build_panel_to_top": false,
 
-  // OPTION: "hide_build_panel"
-  // Controls when the LaTeX build panel is hidden after a build finishes.
-  // Default: "no_warnings"
-  //
-  // Possible values are:
-  //   "always"      — always hide the panel, even if the build failed.
-  //   "no_errors"   — hide the panel if there are no errors.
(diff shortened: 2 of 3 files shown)


deathaxe:
Thanks.

Actually the whole setting is mis-leading as it doesn't control whether build panel is hidden, but whether it is displayed. An already open panel is not actively hidden.

alberti42:
Yes, I totally agree with you. Ideally, we should look for a better name, perhaps deprecating the old one. I would suggest the following:

```text
// OPTION: "automatically_open_build_panel"
// Controls when the LaTeX build panel is opened automatically after a build finishes.
// Default: "on_badboxes"
//
// Possible values are:
//   "always"      — always open the panel, even if the build was skipped.
//   "on_badboxes" — open the panel if there are badboxes, warnings, or errors
//                   (when badbox reporting is enabled).
//   "on_warnings" — open the panel if there are warnings or errors.
//   "on_errors"   — open the panel only if there are errors.
//   "never"       — never open the panel automatically. The panel can still be opened manually.
"automatically_open_build_panel": "on_badboxes",
```

which makes the intent much clearer and mirrors what the code does.

deathaxe:
I had default setting's name `"show_panel_on_build": true` in mind. It would then show as `latextools.show_panel_on_build` in project-specific settings.

alberti42:
.. Just a few thoughts... I would very much like to keep the current functionality, namely the possibility to open the build panel only when errors are generated, or only when warnings and errors, etc. I think it would be a pity to reduce this option to a true/false because LaTeX compilation is more nuanced than compiling regular programs. For example, I often do not want to get distracted by many warnings that are generated and that I can review a second time, like missing labels, etc., but I want to focus on what I am writing. This is why I thought this option is great, where we can distinguish so many different use cases.

About the wording: `show_panel_on_build` is nice, short, and clear. But it does not reflect what the current code is doing. I like very much what the code is doing, so I would rather bend the description to the code. A more faithful description would be `open_panel_on_build`. So, if the panel was already opened and the user chose `never`, it will still remain visible without contradicting the label.

deathaxe:
My post was about naming, only - not about changing functionality. The setting would use your values "on_error", ... .

alberti42:
I was probably confused by `true` as a possible value. Do you plan to also include `true` and `false` as possible logical values? This would probably be confusing. The name `show_panel_on_build` is better than the current one. However, I would find `open_panel_on_build` even better for the reasons above. In any case, a massive thank you for maintaining this indispensable plugin!

Issue #1680, opened 2025-10-10
two (related?) issues with build paned and error handling


pierredelun3:
This is a bit strange, since this problematic behaiour disappeared after a clean reinstall of ST and then reappeared again. Two problems which are maybe related, that's why I put them together.
Build 4200 on Silicon Mac 15.6.1, Latextools 4.5.5

1) After building a latex file (whether or not the compilation produces errors), if I issue from ST menu the command "Show Results Panel", an empty panel is shown. If the Latextools panes was open, the empty panel is superimposed to it. Is this the intended behaviour? Shouldn't it be Latextool Build Panel = Results panel = output.exec ?
In any case, giving in the console the command
   window.run_command("show_panel", {"panel": "output.exec"})
fails silently

2)  If the compilation of a latex file produces errors, the build panel is open,  errors are signaled, but the wrong file path is shown, pointing at

/usr/local/texlive/2025/texmf-dist/tex/latex/l3kernel/expl3.sty:179: Undefined control sequence. [\aaa]

The line number is correct (in this case, an undefined control sequence \aaa at line 179), but the path points at a system file. The command Next Result then opens the expl3.sty file at line 179.

Can anyone point me at the right direction to debug this?

deathaxe:
For the answer, refer to https://forum.sublimetext.com/t/issues-with-latextools-build-panel/77262/2?u=deathaxe

PR #1681, opened 2025-10-10
Use ST's default build output panel

deathaxe:
Addresses #1680

This commit teaches make_pdf command to re-use ST's default build output panel "output.exec" to integrate more seamlessly.

1. It enables displaying build panel via default main menu item

   Menu > Tools > Build Results > Show Build Results

2. It makes separate key bindings for `shift+escape` obsolete

### DIFF
--- a/Default (Linux).sublime-keymap
+++ b/Default (Linux).sublime-keymap
@@ -642,16 +642,6 @@ LaTeX Package keymap for Linux
 		],
 	},
 
-	// Show panel
-	{
-		"keys": ["shift+escape"],
-		"command": "show_panel", "args": { "panel": "output.latextools" },
-		"context": [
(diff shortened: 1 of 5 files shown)


michaelblyons:
thanks. the first time i used lt, this confused me, too.

do the "show output when..." options still work? or is there some st default that messes with it?

deathaxe:
This PR just renames the created panel name. It causes default `exec` and `make_pdf` to share same panel control. Beyond that, nothing changes.

ST's default setting to control build output is `"show_panel_on_build"`. As a boolean setting, it is insufficient, thus LaTeXTools keeps using `"latextools.show_panel_on_build"` to provide more fine gained control about, when to display build output.

PR #1688, opened 2025-11-24
Autodetect latex log file encoding

### LINKED ISSUE #1687: LaTeXTools could not parse the TeX log file
This error is new (perhaps a week). Version 4.5.7, Windows 11, latest MiKTeX

LaTeXTools could not parse the TeX log file C:\Users\...\Ch_16_Fin2Bond1.log
(actually, we never should have gotten here)

Python exception: UnicodeDecodeError('utf-8', b"This is pdfTeX,

deathaxe:
Fixes #1687

This PR addresses behavior of 

1. pdfLaTeX, which creates ANSI encoded logfiles, likely using OS's default encoding (at least on Windows)
2. LuaLaTeX/XeLaTeX, which create utf-8 encoded log files.
 
There maybe other factors, hence use charset_normalizer package to auto-detect file encoding.

Additional, decoding errors are ignored. That's what the previous logfile parser did, too.

### DIFF
--- a/latextools/utils/tex_log.py
+++ b/latextools/utils/tex_log.py
@@ -1,9 +1,12 @@
 from __future__ import annotations
 from itertools import chain
+import codecs
 import os
 import regex as re
 import sublime
 import sublime_plugin
+
+from ...vendor.charset_normalizer import from_bytes as charset_from_bytes
 
 
 class LogRule:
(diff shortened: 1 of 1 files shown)


zovcin:
The compiling process reports error even after update. 
"
[Compiling 'document.tex' with 'Basic Builder']
running pdflatex...error
"
Second time the error remains, but the pdf get opened by the viewer
So, not fixed completely IMO

deathaxe:
This issue is about produced error log not being parsed due to asssuming wrong file encoding.

> running pdflatex...error

The `error` indicates pdflatex not returning success, when compiling PDF document. That's unrelated and not neccessarily be caused by LaTeXTools. It is just the compiler, which returns non-zero exit code. There are plenty possibilities, which may cause it to do so, all of which indicate the produced document probably not being complete.

PR #1696, opened 2026-03-12
Add autoclosing of an environment when using the latex auto fill feature when approiate

dam024:
This modification improves the Fill-helper autocompletion features of LaTeXTools: 
- When using the fill-helper to autocomplete an environment (\begin{), the environment is automatically closed when appropriate and the cursor is moved to the appropriate position. 
- When fill-helper completes another command than an environment, the cursor is moved after the closing bracket when appropriate
- Settings were made to control the behavior of these new features (please, refer to these for more information)

### DIFF
--- a/LaTeXTools.sublime-settings
+++ b/LaTeXTools.sublime-settings
@@ -101,6 +101,11 @@
 	// You can also use toggle: C-l,t,a,e
 	"env_auto_trigger": false,
 
+	// Fill-helper autocompletion will close automatically an environment after \begin{. This trigger is only ran after you have selected an environment from the list
+	// This autoclose is only triggered when adding a new \begin{, not when modifying an existing one via the Window Command Palette
+	// This requires "env_auto_trigger" to be true. 
+	"env_autoclose_trigger": false,
+
 	// Fill-helper autocompletion triggered for a wide range of references to external
 	// files. You can also use toggle: C-l,t,a,f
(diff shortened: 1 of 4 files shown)


deathaxe:
1. `"smart_cursor_move_auto_trigger": true` doesn't seem to have any effect on my setup. Caret is not moved after autofill helper inserts environment name.

2. With `"env_autoclose_trigger": true` corresponding `\end` command is unconditionally inserted, regardless a matching end command being present in the document. While this may help when adding new environments, it may be unwanted when just editing/changing existing ones.

   Just consider someone navigating into existing \begin{...}, hitting `ctrl+l,ctrl+shift+n` to change the name via autofill helper. With this PR 2 (broken) end commands are inserted.

3. Behaviour of committing environment names via auto-completions and auto-fill helper should more or less be the same for consistency reasons. I consider auto-fill helper as a legacy mechanism to provide completions which dates back to days ST's auto-completions didn't exist or were not powerful enough.

3. There are already various ways to insert matching ´\begin..\end´ commands:
   - via plain \begin snippet 
   - via CWL completions
   - via <kbd>ctrl+l,ctrl+n</kbd>

Regardless the function being gated behind settings, it should behave in sane and consistent ways, when enabled. Otherwise it won't be used by many or will just be cause of follow up issues and/or unnecessary extra maintenance burden.

